PR #31, opened 2022-01-28
Salty Runback & Quick Exit (Wifi-Compatible)

### LINKED ISSUE #28: Salty Runback + Quick Exit Implementation
LRASTART should be a button combo that checks to be held instead of recursively pressed, this allows you to ergonomically do the combo faster in one motion, and without worrying about order, this is something pm has and it feels really nice

along with that, the yes no prompt should either immediately open, or be skipped altogether
imo please skip it altogether once the button combo held check mentioned above is true

ergonomic extra button held for different actions, as follows:
* LRASTART   - no contest screen, no prompt to confirm Yes/No
* LRASTART+X - quick exit, go back to sss
* LRASTART+Y - salty runback

blu-dev:
Resolves #28 aside from LRASTART without X+Y being exit to No Contest.

### DIFF
--- a/fighters/mario/src/opff.rs
+++ b/fighters/mario/src/opff.rs
@@ -213,12 +213,6 @@ pub unsafe fn moveset(fighter: &mut L2CFighterCommon, boma: &mut BattleObjectMod
     noknok_timer(fighter, boma, id);
     noknok_reset(fighter, id, status_kind);
     noknok_training(fighter, id, status_kind);
-
-    if fighter.is_button_on(Buttons::AppealHi) {
-        utils::util::trigger_match_reset();
-    } else if fighter.is_button_on(Buttons::AppealLw) {
-        utils::util::trigger_match_exit();
-    }
 }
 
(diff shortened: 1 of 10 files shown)


fighters/mario/src/opff.rs
```diff
@@ -213,6 +213,12 @@ pub unsafe fn moveset(fighter: &mut L2CFighterCommon, boma: &mut BattleObjectMod
     noknok_timer(fighter, boma, id);
     noknok_reset(fighter, id, status_kind);
     noknok_training(fighter, id, status_kind);
+
+    if fighter.is_button_on(Buttons::AppealHi) {
```
techyCoder81: hey actually why is this here?

techyCoder81 (changes requested):

fighters/mario/src/opff.rs
blu-dev: Accident left in for when I was testing. Has been resolved.

techyCoder81 (approved):

PR #54, opened 2022-01-29
install hook on arcrop hook for version string (#35)

### LINKED ISSUE #35: hdr version string in main splash screen ui
add the hdr version string to the splash screen, with arcropolis, etc.

related to #24

blu-dev:
Resolves #35 by hooking ARCropolis' change_version_hook, which is fine since we have a hard dependency on ARCropolis anyways

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
 
     # build the project
     - run: |
-        PATH=$PATH:/root/.cargo/bin && HDR_VERSION="${{ steps.version.outputs.version_string }}" && cd scripts && python3 make_dist.py release && cd ..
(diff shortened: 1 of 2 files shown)


WuBoytH (changes requested):
`"{}\nHDR Ver. {}\0"` will create a third line and will cut off the smash version. I'd suggest changing to `"{}, HDR Ver. {}\0"` (both smash version and HDR version on the same line) so that nothing gets cut off on a vanilla title screen layout.arc.

blu-dev:
image

WuBoytH (approved):
Looks good to me, lgtm
ofc this is subject to change if we make a title screen layout.arc but for now this is perfect

PR #58, opened 2022-01-30
Green Missile Rework

### LINKED ISSUE #56: Give Luigi a green flash when misfire is on deck and let him store it
Luigi's misfire should flash green when it's in the charge phase, and then flash a green color when he has it. If you press shield he should store the charge for an increasingly lower multiplier down to a minimum of 50%.

### LINKED ISSUE #57: vars::roy::TRAIL_EFFECT is turned off on death even if not chroy
Will lead to unexpected behavior on any other character who has a flag set like that that needs to persist over death

blu-dev:
This resolves #56 and resolves #57 

Basically, gives Luigi a better/more predictable misfire and misfire storage.

Currently, Luigi's misfire is calculated at the end of the charge, before the launch, with a probability of 1/8.

However, you were not guaranteed to get a misfire once out of every 8 usages. Now, the misfire has the following behavior:
* At the beginning of a match (or a training mode reset), a random number in the range [0, 7] (inclusive) is generated. This number designates how many more Green Missiles he must perform before getting a misfire. Misfires that go past their time limit and result in Luigi panting also count towards this number.
* When Luigi is going to perform a misfire, the charge effects will be green, as shown:
image
* During a misfire charge, if you press either guard (shield), or guard hold (shield lock), the misfire will be stored and the effects will be replaced by regular charge effects. Every time you store a misfire, the damage multiplier for misfire is decreased by the following formula: `max_damage = 25.0 * max(0.95^x, 0.8)`, where `x` is the number of times you have stored the same misfire, and `0.8` is the lower bound, meaning you can never go below `80%` of the max damage (25.0). This is also scaled later on by how long you held the misfire for.
* Misfire storage lasts across deaths.
* Misfire has green effects for all of the missile part of the move, including green smoke trail
* The green effects will show for a minimum of 3 frames. After the third frame, if you are pressing shield then it will be stored. This cannot be accidentally tap buffered, it checks the raw button input after the third frame.
* Here is a video in action:

https://user-images.githubusercontent.com/42755519/151686799-6c46ea66-e3ba-4936-818c-ecb60da5c24b.mp4

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -223,6 +223,24 @@ pub mod vars {
         pub use super::shotos::*;
     }
 
+    pub mod luigi {
+        // flag
+        /// This flag stores whether or not Luigi currently has a misfire stored.
+        pub const IS_MISFIRE_STORED: i32 = 0x1000;
+        
+        // int
+        /// This int stores the number of remaining green missile's luigi must do before getting a misfire
+        pub const REMAINING_SPECIAL_S_UNTIL_MISFIRE: i32 = 0x1000;
+        /// This int stores the handle of the charge smoke effect for killing it if we store misfire
+        pub const CHARGE_SMOKE_EFFECT_HANDLE: i32 = 0x1001;
+        /// This int stores the handle of the pulsing effect for killing it if we store misfire
+        pub const CHARGE_PULSE_EFFECT_HANDLE: i32 = 0x1002;
+
+        // float
+        /// This float holds the current multiplier on damage for misfire
+        pub const MISFIRE_DAMAGE_MULTIPLIER: i32 = 0x1000;
+    }
+
     pub mod mario {
         // flags
         pub const FIREBRAND_SPAWNED: i32 = 0x1000;
--- a/fighters/common/src/function_hooks/init_settings.rs
+++ b/fighters/common/src/function_hooks/init_settings.rs
@@ -107,7 +107,8 @@ unsafe fn init_settings_hook(boma: &mut BattleObjectModuleAccessor, situation: s
         */
 
         //Sword trails
-        if VarModule::is_flag(boma.object(), vars::roy::TRAIL_EFFECT) {
+        if (boma.kind() == *FIGHTER_KIND_ROY || boma.kind() == *FIGHTER_KIND_CHROM) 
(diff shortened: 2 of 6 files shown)


JoaoAlmeida20:
Everything seems to be working as described, tested every detail listed including training mode resets the random number, luigi panting counting towards the counter, damage decay including the lower bound, storage lasting across deaths.
Animation looks a bit off if you put off cancelling a misfire charge until near the end where luigi is about to give up, as the animation sort of resets for a short bit to the standard charging part but then jumps back to him giving up quickly. Seems to have about the same timing it would have normally and works fine functionally so just a small note

blu-dev:
Thanks for the feedback. I forgot to remove the effect at the right time, so I will resolve that in a little bit.

techyCoder81:
Leaving this as a general comment (havent had time to actually do a review yet), but I wonder if it would be cool to add in ParamModule an entry for the "effect color" of each character's skin. So like, for example, luigi would have 8 values in a "colors" field that indicates what color any of his special effects should be based on skin slot. That would let us specify things like this, or like inkling's paint colors, in a more clean manner that would be consistent across characters. We could just call something like `fighter.get_skin_color` or `fighter.get_effect_color` to access it cleanly. Then any vfx we make will use those values. Might be a lot of work, might be worth it. Could always start with python-script-generated white vfx for all characters, and then when characters use them (like luigi or inkling) we can set real values. Users can change them if they have their own colors. Or we can make a special colors.json equivalent for this.

black-calculus (approved):
lgtm

blu-dev:
> Leaving this as a general comment (havent had time to actually do a review yet), but I wonder if it would be cool to add in ParamModule an entry for the "effect color" of each character's skin. So like, for example, luigi would have 8 values in a "colors" field that indicates what color any of his special effects should be based on skin slot. That would let us specify things like this, or like inkling's paint colors, in a more clean manner that would be consistent across characters. We could just call something like `fighter.get_skin_color` or `fighter.get_effect_color` to access it cleanly. Then any vfx we make will use those values. Might be a lot of work, might be worth it. Could always start with python-script-generated white vfx for all characters, and then when characters use them (like luigi or inkling) we can set real values. Users can change them if they have their own colors. Or we can make a special colors.json equivalent for this.

I don't really think that this is a worth while idea, since most people will be playing with cosmetic mods anyways.

Issue #17, opened 2022-01-05
ingame update prompt


techyCoder81:
ingame update prompt for ryujinx

Riah321:
I'm guessing this has been addressed. If so please close it :)

PR #64, opened 2022-01-31
Automatic Nightly/Beta Updater

blu-dev:
This PR works towards resolving #17 for on-console updaters. For Ryujinx players, there will likely be no update prompts at all because we auto-detect and skip.

image

Afterwards, it will prompt a relaunch of the application where you have to re-select the user

Note: This creates a dependency on devkitpro being installed and added to your path in order to build HDR

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
 
     # build the project
     - run: |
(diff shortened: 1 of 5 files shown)


WuBoytH (approved):
Gotta see how it works later on down the road but for now lgtm

PR #72, opened 2022-02-01
Fix motion-rated DJC override aerial heights

### LINKED ISSUE #27: Motion rated aerials affect DJC override height
replace FT_MOTION_RATE with MotionModule::set_rate calls in aerials.rs

https://user-images.githubusercontent.com/47401664/151902707-a8d6ce3b-1446-4ee2-b0ce-cf07509fb4ac.mp4

jobrien97:
Replaced FT_MOTION_RATE calls with MotionModule::set_rate

Fixes #27


blu-dev (approved):
lgtm, haven't had the chance to playtest it but looks good.

Can you confirm if the aerials which we don't have reimplemented are also affected by this and will need to be changed later on or are these the only problematic ones?

jobrien97:
> lgtm, haven't had the chance to playtest it but looks good.
> 
> Can you confirm if the aerials which we don't have reimplemented are also affected by this and will need to be changed later on or are these the only problematic ones?

We've reimplemented all DJC characters' aerials, so this should cover it all

PR #81, opened 2022-02-02
Fix turndash cancelable moves triggering from turn input

### LINKED ISSUE #80: Turndash cancels trigger from turn input
Replace Turn cat checks with TurnDash

jobrien97:
replace Turn cat checks with TurnDash

Fixes #80


blu-dev (approved):
lgtm

PR #83, opened 2022-02-02
Fix being able to jump out of shield lock with tap jump

### LINKED ISSUE #82: Can jump out of shield lock with tap jump
Replace Jump cat check with JumpButton

jobrien97:
replaced Jump cat check with JumpButton

Fixes #82


blu-dev (approved):
lgtm, just update to dev

PR #87, opened 2022-02-04
Fix plat drops out of dash/run

### LINKED ISSUE #60: Dropping through platforms while running doesn't work

black-calculus:
Changed the is_status method to change_status_req, incorrect method was used there

Resolves #60


blu-dev (approved):
lgtm, dunno how that happened

PR #95, opened 2022-02-05
set roll dir in VarModule for glide toss (#61)

### LINKED ISSUE #61: Glide Tossing (Roll cancelled item throws) don't work
You can cancel roll with an item throw but you stay stationary/practically stationary, instead of keeping the roll momentum during the item throw

blu-dev:
Resolves #61 by putting back the line that sets the roll direction in glide toss


black-calculus (approved):
lgtm

PR #96, opened 2022-02-05
add dacus params and fix dacus

### LINKED ISSUE #59: DACUS/DACDS doesn't work
Dash attack is not cancellable into neither up smash nor down smash, tested input as fast as possible in frame by frame. Input method of just using the same attack button that was used for the dash attack and using the main stick for the direction

blu-dev:
Resolves #59 

Adds a new param struct and two new params:
```xml
<struct hash="dacus_enable">
    <int hash="start_frame">3</int>
    <int hash="end_frame">10</int>
</struct>
```

Dacus is based off of these params now and the ints are status frames


WuBoytH (changes requested):
For whatever reason, DACUS/DACDS cancel *actually* happens from frames 5 to 12 (tested with Luigi). There might be some logic that needs to be reviewed.

WuBoytH (approved):
seems to work fine now, lgtm

PR #100, opened 2022-02-06
fix agent params in parammodule

### LINKED ISSUE #90: ParamModule does not handle agent params correctly
Ideally I'd like to handle this so that Luigi misfire parameters can be stored in ParamModule.

blu-dev:
Resolves #90 and also adds Luigi params for misfire storage/green missile rework


WuBoytH (changes requested):
There's a crash when loading the params, it seems. Outside of that, the code itself looks good, but I'm gonna leave it at "Request changes" until that issue is resolved.

WuBoytH (approved):

PR #102, opened 2022-02-07
Fix crouch out of run/run brake

### LINKED ISSUE #101: Crouching out of run/run brake doesn't work
https://github.com/HDR-Development/HewDraw-Remix/blob/917b19de4f83460dcec346cc320752a9c174dbc8/fighters/common/src/opff/tech.rs#L183

change_status_req is trying to change to idle instead of squat

JoaoAlmeida20:
change_status_req was changing/(trying to change?) to idle instead of crouch. tested and works now

fixes #101


WuBoytH (approved):
Looks good to me, let's get this money

JoaoAlmeida20:
https://streamable.com/f9p0r4 footage of fix

PR #112, opened 2022-02-14
It's Gone, Rejoice

### LINKED ISSUE #111: Disable Kill Zoom
Needs to NOP instructions at the following offsets: `0x637384`, `0x633dc0`

The former is for disabling the throw slowdown and the latter is for disabling the regular hitlag

blu-dev:
Kill zoom is gone, resolves #111 

Previously there were issues because instead of preventing the actual "finish" state from triggering, we just killed the finish cut in, which left hitlag. This removes the hitlag, slow down on throws, and the finish zoom


WuBoytH (approved):
Looks good to me, let's get this money

PR #116, opened 2022-02-15
small fix for beta releases

techyCoder81:
beta releases were not created correctly after PR merge because commitish is not a valid option for this action.


blu-dev (approved):
ligma

PR #133, opened 2022-02-17
Correctly install ICs opff (fixes crashes)

### LINKED ISSUE #131: Nana tech not working/crashing
- Cheer cancels
- Infinite hold buffer
- Death gfx crashing

jobrien97:
Fixes #131


techyCoder81 (approved):
I would love to see this done using `boma->partner` equivalents instead of using ID and a local static array, because we know `BattleObjectModuleAccessor` has a `partner` field that the game uses in ICs status scripts to get nana from popo, and vice versa, but I don't know if thats available to us right now in skyline.

Using VarModule would also be better than a static array, but given that this is a crash fix, imo its a priority change.

techyCoder81:
@blu-dev probably knows how we can get that data - know its been discovered already in ghidra the part of the struct that stores partner, but I can't recall what it was.

PR #151, opened 2022-02-19
(bugfix) Fix ICs jump momentum

### LINKED ISSUE #142: Popo jump momentum bugged
Old bug got reintroduced in refactor

Popo will jump with 0 horizontal velocity after dash dancing; I suspect Popo is using Nana's jump momentum var values

jobrien97:
ICs no longer jump with 0 momentum after dash dancing

Fixes #142


blu-dev (approved):
lgtm

PR #153, opened 2022-02-27
(refactor) tag: player who is "it" turns red when timer is about to end

### LINKED ISSUE #139: change tag gamemode aura to red during last few seconds
if there is only like 3 seconds left before a tag expires, replace the aura with a bright red aura.

relevant info:
`EffectModule::set_rgb_partial_last` - examples can be found by searching the current repo
https://github.com/HDR-Development/HewDraw-Remix/pull/110 - pr that implemented tag
`utils/src/game_modes/tag.rs` - file responsible for handling tag logic
`TagInfo.update_eff()` - potentially call that updates the effect if necessary, logic may belong here

techyCoder81:
player who is "it" now turns red during the last 3 seconds of the timer.

resolves #139


blu-dev (approved):
not a fan of CBM but it's fine, lgtm

PR #178, opened 2022-03-06
(bugfix) resolves vars that are declared and used twice

### LINKED ISSUE #103: Character specific vars are accessed through different modules in opff and acmd, breaking some of their mechanics
Samus's shinespark doesn't work

I think this is because in samus's opff.rs the vars related to shinespark are accessed through the common module, like for example here:
https://github.com/HDR-Development/HewDraw-Remix/blob/dee2134137bc4d375ae899899b23a8064fecb45f/fighters/samus/src/opff.rs#L36

Meanwhile in the acmd script for dash attack the vars are accessed through the samus module, like for example here:
https://github.com/HDR-Development/HewDraw-Remix/blob/dee2134137bc4d375ae899899b23a8064fecb45f/fighters/samus/src/acmd/ground.rs#L12

techyCoder81:
some variables were declared twice, and then both used, where it should have just been one.

resolves #103

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -89,9 +89,7 @@ pub mod vars {
         pub const FOOTSTOOL_AIRDODGE_LOCKOUT: i32 = 31;
         pub const CAN_ESCAPE_TUMBLE: i32 = 32;
         pub const SUPER_CANCEL: i32 = 33;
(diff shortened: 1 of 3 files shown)


blu-dev (approved):
lgtm

PR #244, opened 2022-03-12
(bugfix) tag: disable tag outside regular smash

### LINKED ISSUE #229: Training Mode Tag Mode
If the last rule set selected was using HDR's tag mode, the next time you enter training mode the tag rules will be in effect.

blu-dev:
Resolves #229 

Disables the custom gamemodes outside of any mode that is not match kind 1 (regular smash)


utils/src/offsets.rs
```diff
@@ -358,7 +358,7 @@ mod offsets_impl {
         unsafe {
             let bl = *offset_to_addr::<u32>(bl_offset);
             let imm = bl & 0b0_00000_11111111111111111111111111;
-            bl as usize
+            (imm * 4) as usize
```
techyCoder81, line 361: approving, but also wanting to know what this is for, as it seems unrelated (?). Looks like some register magic, what's it do?

techyCoder81 (approved):

PR #270, opened 2022-03-14
Fix development.nro path in build.py for Ryujinx

Kekker-git:
`development.nro` has to be in the `atmosphere` folder to be modified during runtime. This gets the nro put in the right place.

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -112,7 +112,7 @@ if (is_dev_build and not is_publish):
     print("subpath: " + development_subpath)
     print("type: " + build_type)
     pkgutil.collect_plugin("hdr-switch", os.path.join(switch_rom_path, development_subpath), build_type, "development.nro", "development")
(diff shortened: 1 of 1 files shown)


scripts/build.py
```diff
@@ -112,7 +112,7 @@
     print("subpath: " + development_subpath)
     print("type: " + build_type)
     pkgutil.collect_plugin("hdr-switch", os.path.join(switch_rom_path, development_subpath), build_type, "development.nro", "development")
-    pkgutil.collect_plugin("hdr-ryujinx", os.path.join(ryujinx_rom_path, development_subpath), build_type, "development.nro", "development")
+    pkgutil.collect_plugin("hdr-ryujinx", os.path.join("sdcard", os.path.join(switch_rom_path, development_subpath)), build_type, "development.nro", "development")
```
techyCoder81: this can just be `os.path.join("sdcard", switch_rom_path, development_subpath)` without the nested join

techyCoder81 (changes requested):
quality could be slightly improved

techyCoder81 (approved):

PR #286, opened 2022-03-15
approved duckhunt changes

### LINKED ISSUE #288: approved changes to duck hunt
- [ ] duck hunt should be able to remote detonate gunman. There should be a reasonable frame window (like 20) after inputting the detonation where opponents have time to react to SFX. Otherwise, it should stay out longer before automatically firing.
- [ ] gunman faf could be slightly improved as a result
- [ ] down smash should be individual launching hits, rather than a multihit (which will often fail in this engine)
- [ ] fsmash should be individual launching hitboxes for the same reason
- [ ] clay pigeon does not combo into detonation hardly at all, this could be fixed

elias-michaias:
- [x] Down Air 
 * Air-only BKB 10 > 30
 * Air-only KBG 80 > 60
- [x] Jab 2 
 * Angle 361 > 70
- [x] Side Special 
 * Initial contact hitlag 2 > 5
- [x] Down Special 
 * Shot timer set to 5 seconds, uniform for all gunmen
 * Can be remotely detonated with a down special input
 * Duck Hunt does not enter spawn animation while detonating
 * Added 3 second timer to charge next gunman
 * Flashes white when charged
 * Post-shot frames 100 > 130
 * Health 5 > 7
 * FAF decreased
- [x] Down Throw
 * BKB 70 > 110
 * KBG 50 > 40
- [x] Smash Attacks
 * Added fire effect on hit
- [x] Dash Attack
 * Added reverse property on back side of late hit

Fixes #288

### DIFF
--- a/fighters/common/src/function_hooks/transition.rs
+++ b/fighters/common/src/function_hooks/transition.rs
@@ -165,7 +165,6 @@ unsafe fn is_enable_transition_term_hook(boma: &mut BattleObjectModuleAccessor, 
             && flag == *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_SPECIAL_LW {
                     return false
             }
-            println!("TIMER: {}", VarModule::get_int(boma.object(), vars::duckhunt::GUNMAN_TIMER));
         }
     }   
     original!()(boma, flag)
--- a/fighters/common/src/function_hooks/transition.rs
+++ b/fighters/common/src/function_hooks/transition.rs
@@ -159,13 +159,15 @@ unsafe fn is_enable_transition_term_hook(boma: &mut BattleObjectModuleAccessor, 
             }
         }
 
-        //Disable Duck Hunt Gunman if Gunman is present
-        if boma.kind() == *FIGHTER_KIND_DUCKHUNT 
-        && WorkModule::is_flag(boma, *FIGHTER_DUCKHUNT_INSTANCE_WORK_ID_FLAG_GUNMAN_OK) 
-        && flag == *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_SPECIAL_LW {
-            return false;
-        }       
-    }
+        //Disable Duck Hunt Down Special on a timer
+        if boma.kind() == *FIGHTER_KIND_DUCKHUNT  {
+            if VarModule::get_int(boma.object(), vars::duckhunt::GUNMAN_TIMER) != 0 
+            && flag == *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_SPECIAL_LW {
+                    return false
+            }
+            println!("TIMER: {}", VarModule::get_int(boma.object(), vars::duckhunt::GUNMAN_TIMER));
+        }
+    }   
     original!()(boma, flag)
 }
 
--- a/fighters/common/src/function_hooks/transition.rs
+++ b/fighters/common/src/function_hooks/transition.rs
@@ -158,6 +158,13 @@ unsafe fn is_enable_transition_term_hook(boma: &mut BattleObjectModuleAccessor, 
                 }
             }
         }
+
+        //Disable Duck Hunt Gunman if Gunman is present
+        if boma.kind() == *FIGHTER_KIND_DUCKHUNT 
+        && WorkModule::is_flag(boma, *FIGHTER_DUCKHUNT_INSTANCE_WORK_ID_FLAG_GUNMAN_OK) 
+        && flag == *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_SPECIAL_LW {
+            return false;
+        }       
     }
     original!()(boma, flag)
 }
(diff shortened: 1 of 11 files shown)


JoaoAlmeida20:
everything seems to be working nicely

fighters/common/src/function_hooks/transition.rs
```diff
@@ -158,7 +158,16 @@ unsafe fn is_enable_transition_term_hook(boma: &mut BattleObjectModuleAccessor,
...
         }
-    }
+
+        //Disable Duck Hunt Down Special on a timer
+        if boma.kind() == *FIGHTER_KIND_DUCKHUNT  {
+            if VarModule::get_int(boma.object(), vars::duckhunt::GUNMAN_TIMER) != 0 
+            && flag == *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_SPECIAL_LW {
+                    return false
+            }
+            println!("TIMER: {}", VarModule::get_int(boma.object(), vars::duckhunt::GUNMAN_TIMER));
```
techyCoder81: please comment out this println, we dont want them there in real builds

fighters/duckhunt/src/opff.rs
```diff
@@ -4,44 +4,130 @@ use super::*;
...
 
  
-unsafe fn duck_jump_cancel(boma: &mut BattleObjectModuleAccessor, status_kind: i32, cat1: i32, frame: f32) {
-    if status_kind == *FIGHTER_DUCKHUNT_STATUS_KIND_SPECIAL_HI_FLY {
-        if frame > 20.0 {
-            if boma.is_cat_flag(Cat1::SpecialHi) {
-                StatusModule::change_status_request_from_script(boma, *FIGHTER_DUCKHUNT_STATUS_KIND_SPECIAL_HI_END, true);
-            }
-        }
+unsafe fn duck_jump_cancel(fighter: &mut L2CFighterCommon) {
```
techyCoder81, line 7: this isnt a jump cancel, so it would be nice to rename it to just up special cancel or something. Also, it looks like you can only cancel this with up special (meaning you cant just cancel it with *any* special input). Is this intended?

fighters/duckhunt/src/opff.rs
```diff
@@ -4,44 +4,130 @@ use super::*;
...
+    && fighter.motion_frame() > 20.0
+    && fighter.is_cat_flag(Cat1::SpecialHi) {
+        fighter.change_status_req(*FIGHTER_DUCKHUNT_STATUS_KIND_SPECIAL_HI_END, true);
     }
 }
 
-pub unsafe fn moveset(boma: &mut BattleObjectModuleAccessor, id: usize, cat: [i32 ; 4], status_kind: i32, situation_kind: i32, motion_kind: u64, stick_x: f32, stick_y: f32, facing: f32, frame: f32) {
-    duck_jump_cancel(boma, status_kind, cat[0], frame);
+unsafe fn frame_data(fighter: &mut L2CFighterCommon) {
+    if fighter.is_status(*FIGHTER_STATUS_KIND_SPECIAL_N) {
```
techyCoder81, line 16: i know i changed the calls for these rq and gave them to you, but they could be very slightly improved with a match. not really important though

fighters/duckhunt/src/opff.rs
```diff
@@ -4,44 +4,130 @@ use super::*;
...
+            return
+        }
+        let owner_id = WorkModule::get_int(weapon.module_accessor, *WEAPON_INSTANCE_WORK_ID_INT_LINK_OWNER) as u32;
+        if weapon.is_status(*WEAPON_DUCKHUNT_GUNMAN_STATUS_KIND_READY) {
+            let duckhunt = utils::util::get_battle_object_from_id(owner_id);
+            let duckhunt_boma = &mut *(*duckhunt).module_accessor;
+            if duckhunt_boma.is_cat_flag(Cat1::SpecialLw) && duckhunt_boma.is_button_trigger(Buttons::Special | Buttons::SpecialRaw) && WorkModule::get_int(weapon.module_accessor, *WEAPON_INSTANCE_WORK_ID_INT_LIFE) > 25 {
+                PLAY_STATUS(weapon, Hash40::new("se_duckhunt_special_l09"));
+                let gunman_kind = WorkModule::get_int(weapon.boma(), *WEAPON_DUCKHUNT_GUNMAN_INSTANCE_WORK_ID_KIND);
+                if PostureModule::lr(weapon.boma()) == -1.0 {
```
techyCoder81: instead of having a huge `if` here with duplicated code (I assume), you could just use the `lr` value in your `EFFECT_FOLLOW` arguments and it would be a lot simpler and cleaner.

fighters/duckhunt/src/opff.rs
```diff
@@ -4,44 +4,130 @@ use super::*;
...
+                        }
+                        _ => {
+                            return
+                        }
+                    }
+                }
+                if PostureModule::lr(weapon.boma()) == 1.0 {
+                    match gunman_kind {
+                        0 => {
+                            EFFECT_FOLLOW(weapon, Hash40::new("duckhunt_wildegunman_light"), Hash40::new("top"), -0.5, 13.3, 0.74, 0, 0, 0, 1, true);
```
techyCoder81: for example, all of these `-0.5` values could just be `0.5 * lr`

techyCoder81 (changes requested):
some relatively small stuff could be cleaned up overall

fighters/duckhunt/src/opff.rs
elias-michaias: Calc put that there, I have not touched it. And yes, basically you can cancel into a move, or you can cancel with an up special input in order to go into freefall immediately (intentionally) and grab ledge. This is already HDR behavior. Without it, his recovery would be dogwater as there is no ledgesnapping in this game.

fighters/duckhunt/src/opff.rs
elias-michaias: https://doc.rust-lang.org/1.6.0/book/patterns.html#guards
I think it would become more verbose. I feel like leaving at two IFs is more readable personally. I would switch to a match with 3 values or more. If you want to change it though, I'm fine.

techyCoder81 (approved):
lgtm

PR #287, opened 2022-03-15
Fix double traction

### LINKED ISSUE #218: traction is too slippery
need to look into why traction is universally so slippery and hard to control

jobrien97:
Double traction is now properly applied when you are above max walk speed (during certain states)

Needs to be coupled with fighter_param_motion.prc changes HDR-Development/hdr-private/pull/35

Fixes #218


techyCoder81 (approved):
looks reasonable enough to me, lets see how it feels in nightly/beta

PR #320, opened 2022-03-17
Fix double traction during sliding attacks

### LINKED ISSUE #306: Bug: attacks moonwalking you backwards
tested on nightly 24 and 26, cloud, joker, kazuya and the belmonts' down tilt, and bayonetta's side tilt move the character back during the move in a strange way, it's fun to play with, but you devs decide if this should be in or not. this happened on linux (arch) with high end amd computer specs

https://user-images.githubusercontent.com/33030702/158655155-b5aef24a-1fdd-4437-a207-c92d17eb07f0.mov

.

### LINKED ISSUE #313: Moves that change the fighter's position go backwards
Version: v0.3.26
RomFS Version: v0.10.8
Platform: Ryujinx Skyline PR
Certain moves that move the fighter will start moving the fighter in the opposite direction near the end of the move. 
Examples include Cloud's D-Tilt, Kazuya's F-smash, Joker D-tilt, Mega Man D-Tilt

https://user-images.githubusercontent.com/62036967/158707089-012c7715-1038-4d88-94c4-6f3823734a93.mp4

It seems like any move that moves the "Trans" bone has this issue.

jobrien97:
Double traction is no longer active during motion-based attacks (moves that change your position)

Fixes #313
Fixes #306

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -112,7 +112,7 @@ pub mod vars {
         pub const IS_MOONWALK: i32 = 57;
         pub const CAN_GLIDE_TOSS: i32 = 58;
         pub const IS_MOONWALK_JUMP: i32 = 59;
-        pub const ENABLE_DOUBLE_TRACTION: i32 = 60;
+        pub const IS_MOTION_BASED_ATTACK: i32 = 60;
         pub const PREV_FLAG_DISABLE_ESCAPE_AIR: i32 = 61;
         pub const ENABLE_WAVELAND_PLATDROP: i32 = 62;
         pub const SPECIAL_PROJECTILE_SPAWNED: i32 = 63;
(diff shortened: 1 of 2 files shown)


techyCoder81 (approved):
approving for now, but not a fan of this impl. Looking forward to when we can do it right, from the bottom up, rather than from the top down.

techyCoder81:
there are reports that this is not resolving the issue for certain moves in the #testing channel

techyCoder81 (changes requested):
there are reports that it still has edge cases and does not work uniformly. Shouldnt we be checking kinetic type or something instead?

techyCoder81:
made a new ticket to track the fact that this still needs work. However, merging these changes because they will at least improve the current problems.

PR #329, opened 2022-03-18
fixing rate limits on pr label changes by using release token

techyCoder81:
pr label validation uses the github token, which means it is getting rate limited. This change switches to the release token (which is a personal access token) to avoid rate limits that are stopping pr validation. We also switch to using PR target, so that people cannot abuse this as a security concern.

### DIFF
--- a/.github/workflows/semver_label_check.yml
+++ b/.github/workflows/semver_label_check.yml
@@ -1,6 +1,6 @@
 name: validate pr labels
 on:
-  pull_request:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


zandm7:
Just FYI this (via v33 nightly) seems to crash Ryujinx upon loading into a match/training.

Not sure how/if Switch performance is affected; somebody else chime in?

EDIT: Tested Ryujinx with v30 nightly and the problem is no longer present, so it's either this or #326, I think.

PR #336, opened 2022-03-19
Bowser Jr. - upB adjustments

jobrien97:
- Fix AGT bugs (again)
- Edge cancelable upB landing

Fixes #207


JoaoAlmeida20:
Seems to fix all variations of the AGT bug, and the edge cancelling is working correctly

JoaoAlmeida20:
Unsure if I should make a separate issue for this but it happened in this build even though it's possible it's unrelated:
https://streamable.com/f3igy4

techyCoder81:
if this cannot be fully resolved, we should be considering remove the relevant changes entirely or something. This has been going on for quite a while now. seems like adding more and more logic to catch the edge cases is getting ridiculous.

JoaoAlmeida20:
I do doubt the last bug has to do with the AGT bug and all the relevant changes, but it's hard to say because I haven't been able to replicate it neither in this build nor the nightly. It seems to just have to do with the up b code in general, I feel like I airdodged and on the first frame it snapped me to the ground as if I was right next to it, where the airdodge does snap to the ground purposefully

techyCoder81 (approved):
i guess im just going to approve this because its been sitting for so long. Want to note that there may be more issues here and they should *really* be resolved at some point.

PR #345, opened 2022-03-20
Bowser Jr Up B Explosion adjustments

### LINKED ISSUE #343: Bowser Jr Up B explosion projectile multihit inconsistency
The explosive remains Bowser Jr's up B leaves behind have two hits, a first hit while it's falling, that is supposed to link into the second hit, that is the actual explosion. This is inconsistent in vannila and even worse in HDR. This could be addressed by making the projectile explode on impact with the first hit, which would be more in line with how most explosive projectiles work in this game anyway.

JoaoAlmeida20:
Changes the explosive projectile on jr's up b to transition into its explosion on hit with the first hitbox to fix problems connecting the move.
https://streamable.com/nkcng2

Fixes #343


jobrien97 (dismissed):
try to avoid nested if statements, otherwise lgtm

jobrien97 (approved):
luv u

PR #348, opened 2022-03-20
Jr side b jump transition into real jumpsquat

### LINKED ISSUE #347: jr side b jc transition to jumpsquat instead of side b jump status
change jr's side b so that when you jump out of it you go into jumpsquat instead of the jr side b jump status, this allows sh or fh instead of fixed height side b jump height, cleaner wavedashes, and jc grabs, upsmashes etc

JoaoAlmeida20:
Fixes #347


jobrien97 (approved):
lgtm

PR #354, opened 2022-03-22
dk bair lowered programmatically

### LINKED ISSUE #351: ModelRotationModule impl
Would be good to implement something like a ModelRotationModule to cleanly handle bone rotations and lerping in frame windows for animations like falcon nair, dk bair, incin angled grab, Ganon nair, etc.

techyCoder81:
dk bair is now angled more flat like it has been in previous games. His boxing potential has been missing this tool because it doesnt hit standing opponents out of shorthop, similarly to falcon's same issue

Also introduces an extension trait to set bone rotations easily, instead of a model rotation module.
Resolves  #351

### DIFF
--- a/dynamic/src/ext.rs
+++ b/dynamic/src/ext.rs
@@ -347,6 +347,7 @@ pub trait BomaExt {
     unsafe fn is_flick_y(&mut self, sensitivity: f32) -> bool;
     unsafe fn is_input_jump(&mut self) -> bool;
     unsafe fn get_aerial(&mut self) -> Option<AerialKind>;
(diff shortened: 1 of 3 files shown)


jobrien97:
Why not just do this in anim itself if we already use a custom anim?

jobrien97:
Why not just do this in anim itself if we already use a custom anim?

techyCoder81:
because i didnt wanna make angel have to go adjust it manually, basically

It was also a test for  #351

jobrien97 (approved):
lgtm

PR #370, opened 2022-03-23
Aerial Magic Series now checks Player LR

### LINKED ISSUE #369: Shotos can cancel fair into fair repeatedly when facing left
Magic Series should not allow fair -> fair, yet it is possible...only when facing left

https://user-images.githubusercontent.com/47401664/159734018-de13f89d-10ab-4c8d-b80e-ab01e7215b4d.mp4

WuBoytH:
Resolves #369


techyCoder81 (approved):
looks like a nice minor cleanup to me

PR #371, opened 2022-03-23
IRAR rework v2

jobrien97:
IRAR input:
- IRAR can now only be performed out of a smash turn (in dash, input smash turn, then jump on next frame) ((same input as a pivot jump in Melee/PM))
- Stick must be below dash threshold to apply perfect pivot boost into IRAR. Otherwise, you will pivot jump in place


techyCoder81 (approved):
Good enough for now but may require rework with the new engine

PR #379, opened 2022-03-25
Double traction refactor

jobrien97:
Slight change to accommodate tiny vanilla nuance:
- If you transitioned from one traction-affected status to another (e.g. landing -> crouch), traction gets immediately applied on f1 of crouch 
- However, if you transition from a NON traction-affected status (e.g. run -> crouch), traction does not get applied until frame 2 of crouch (yes this is vanilla behavior)

Also refactored code to be neater


blu-dev (dismissed):
lgtm

blu-dev (approved):

PR #386, opened 2022-03-26
adding new prcxml files as diffed from hdr-assets (not stages)

### LINKED ISSUE #73: Move all PRC files currently in our romfs files over to patch files
With ARCropolis v3.0.0-beta.3, PRC patch files are supported. Currently the recommended format is `.prctxt` so we could package them that way, however in the future it would be wise to implement in the build script an implementation using [prcx](https://github.com/blu-dev/prcx) to convert them too/from

techyCoder81:
needs HDR-Development/hdr-private#51
resolves #73


jobrien97 (approved):
our param changes seem to be intact here on first glance, obviously I cant filter through everything lol, we just need to playtest for a bit ig

techyCoder81:
fwiw, i played with this for 6 hours yesterday and didnt see anything particularly out of order

PR #404, opened 2022-03-29
Samus S4 Reverted Missiles

CodenameComlink:
Hey y'all. This is my first pull request, it's to make Samus missiles work like they did in Smash 4 which should line up better with their Melee counterparts. This is intended to go with the floaty Samus param changes.


techyCoder81:
if this PR is not ready or closed in the next few days I will be closing it as stale.

PR #406, opened 2022-03-29
Reduce parry window 5f -> 3f

### LINKED ISSUE #403: Reduce parry window
5f -> 3f

common.prc

jobrien97:
Param of 2 = ingame 3 frames

Resolves #403


techyCoder81 (approved):
As long as we are sure this is functionally 3 frames, not 2 lol

PR #418, opened 2022-04-01
Jab lock rework

### LINKED ISSUE #258: jab locks adjustments
- [ ] jab locks should not allow getup attack
- [ ] jab lock flop animations should be made much faster
- [ ] when being jab locked, your tech option should be "locked in" by the jab lock

jobrien97:
- Jab lock "flop" animation reduced 40f -> 14f
- Getup attack disabled out of jab lock
- Your getup option is now "locked in" based on your stick position during jab lock hitlag: if your stick is held in a direction during hitlag, you will automatically roll in that direction once your flop animation is over. If your stick is neutral during hitlag, you will automatically regular getup once your flop animation is over. This means that inputting any button, or nothing at all, after being jab locked will still automatically transition you into regular getup.

Resolves #258

### DIFF
--- a/fighters/common/src/general_statuses/downdamage.rs
+++ b/fighters/common/src/general_statuses/downdamage.rs
@@ -1,0 +1,105 @@
+use super::*;
+use globals::*;
+
+// This file contains code for jab locks
+
+pub fn install() {
+    skyline::nro::add_hook(nro_hook);
+    skyline::install_hooks!(
+        get_down_stand_fb_kind_hook
+    );
+}
+
+fn nro_hook(info: &skyline::nro::NroInfo) {
+    if info.name == "common" {
+        skyline::install_hooks!(
+            status_DownDamage_Main
+        );
+    }
+}
+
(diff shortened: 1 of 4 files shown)


fighters/common/src/general_statuses/downdamage.rs
```diff
@@ -0,0 +1,129 @@
...
+                fighter.change_status(
+                    L2CValue::I32(*FIGHTER_STATUS_KIND_DOWN_STAND),
+                    L2CValue::Bool(true)
+                );
+                return 0.into();
+            }
+        }
+        fighter.sub_down_chk_reflect_wall();
+        if !StatusModule::is_changing(fighter.module_accessor) {
+            if (fighter.global_table[SITUATION_KIND] == SITUATION_KIND_GROUND && fighter.global_table[PREV_SITUATION_KIND] == SITUATION_KIND_GROUND)
```
blu-dev, line 81: Could be simplified, but logic is complex

blu-dev (approved):
lgtm, if you could simplify that one block that'd be appreciated but it's one block

blu-dev:
i am merging this because it's submissive and mergeable

PR #420, opened 2022-04-02
Marth tipper changes and active frame changes

blu-dev:
Active frames:

- Dair is now active from frames [6, 10) instead of [5, 11)
    - Dair effect now starts 1 frame before the first hitboxes and changes starts fading shortly after, aligned sword glow to sword
- Uair is now active from frames [5, 10) instead of [5, 15), has been sped up to cover animation frames [5.5, 11.5] during those active frames (covers upper 180 degrees)

Hitbox placements:
- Dtilt:
    - Removed capsule hitbox that was never able to be hit
- Utilt:
    - Removed two low priority capsules on frame 1
    - Swapped priority between arm hitbox (now higher) and hilt hitbox (now lower)
        - This means that spacing is now required to get the move to send up at all, even with the sour upwards hit, otherwise they get hit with sakurai angle
    - Decreased tipper hitbox size 4.5 -> 4.0
    - Changed the angle flipper to only send forward and homogenized all hitboxes -- back hits no longer deal more damage/higher knockback
- Ftilt:
    - Decreased tipper hitbox size 4.5 -> 4.0
- Dsmash:
    - Homogenized position of hitboxes on both front and back swing
    - Decreased sword hitbox sizes 4.0 -> 3.75
- Fsmash:
    - Hitboxes now start on frame 9.3 of the animation (frame 10 in game), which shifts the position to the right of marth's head instead of directly above
    - Increased size of hilt hitbox 3.0 -> 3.5
    - Moved hiltbox further up the sword by 1.1 units to decrease tipper supremacy
- Jab1:
    - Decreased angles on all non-tipper hitboxes by 4, tipper by 5
    - Increased hilt hitbox size 3.0 -> 3.5
    - Moved hiltbox further up the sword by 1.5 units to decrease tipper supremacy, shifted arm hitbox forward by 1.5 units to accommodate
    - Decreased tipper hitbox size 4.0 -> 3.5, shifted up the sword by 0.5 to accommodate

General changes:
- no longer compile `change_version_string_hook` on development plugin
- accept IP address argument for build script to automatically send plugins


black-calculus (approved):
lgtm chieferino

PR #436, opened 2022-04-03
Mk improvements/adjustments

techyCoder81:
- [x] fair effects improved to better match the hitboxes (vfx were the size of the sun)
- [x] fair hitbox made slightly smaller, active frames adjusted, hitlag reduced on first 2 hits
- [x] uair hitbox size reduced, bkb slightly increased, active frames reduced
- [x] dair hitbox active frames reduced, bkb slightly increased, angle flipped to pull in at 60 degrees instead of out at 60 degrees
- [x] dash attack has a sourspot at the end where appropriate
- [x] neutral special can rise appropriately now, horizontal movement reduced

- [x] build script now fails out if the build is not successful, rather than continuing and trying to package.


blu-dev (approved):
lgtm

PR #450, opened 2022-04-06
Taunting from Dash/Run/Runbrake and Teeter Cancelling

### LINKED ISSUE #397: [Enhancement] Allow Taunt out of dash, and to Teeter Cancel Taunts
Currently neither are possible, taunt out of dash and teetering taunts

This is a nice fidget to do in pm with everyone, where you;

dash towards ledge, 
taunt, 
and the teeter will cancel the taunt, 
then you can just dash back, dash to ledge again and cancel dash with taunt again, etc.

GnW (and all characters in their own manner) uses this to spam the bell taunt and its fun, would be a nice port from PM

JoaoAlmeida20:
Taunt from dash, run (edit: and runbrake) and cancel (edit: taunts, dashes and landing animations) at the edge with the teeter animation. This requires the stick's x position to be less than |0.75| and for the player to be under a certain speed.

~~Teeter cancelling is by design left open to other actions than just taunting like in PM, but it doesn't outright interrupt all attacks or anything like that.~~

~~Going to mark as a draft for now because of some late night fixes that could have unseen consequences but should be good
Late night fixes in question were kinetic energy check missing the .abs() so teeter cancelling didn't work on the left side and allowing taunt from run~~

Resolves #397

### DIFF
--- a/fighters/common/src/general_statuses/dash.rs
+++ b/fighters/common/src/general_statuses/dash.rs
@@ -417,6 +417,40 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
         fighter.global_table[PAD_FLAG].get_i32() & *FIGHTER_PAD_FLAG_ATTACK_TRIGGER != 0
     );
 
+    if WorkModule::is_enable_transition_term(fighter.module_accessor, *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_APPEAL_U)
+    && fighter.global_table[CMD_CAT2].get_i32() & *FIGHTER_PAD_CMD_CAT2_FLAG_APPEAL_HI != 0
+    && {
+        fighter.clear_lua_stack();
+        fighter.push_lua_stack(&mut L2CValue::new_int(0x1daca540be));
+        app::sv_battle_object::notify_event_msc_cmd(fighter.lua_state_agent);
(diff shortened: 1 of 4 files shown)


fighters/common/src/general_statuses/dash.rs
```diff
@@ -417,6 +417,40 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
...
+    || fighter.global_table[CMD_CAT2].get_i32() & *FIGHTER_PAD_CMD_CAT2_FLAG_APPEAL_S_R != 0)
+    && {
+        fighter.clear_lua_stack();
+        fighter.push_lua_stack(&mut L2CValue::new_int(0x1daca540be));
+        app::sv_battle_object::notify_event_msc_cmd(fighter.lua_state_agent);
+        fighter.pop_lua_stack(1).get_bool()
+    } {
+        interrupt!(fighter, *FIGHTER_STATUS_KIND_APPEAL, false);
+    }
+
```
jobrien97: Can this not be condensed into 1 block?

JoaoAlmeida20:
Single block taunt checks and shoto backdashes -> taunt are both working, and I've also played around 3 hours with it on Random vs Random CPU with no crashes, and every character (a significant portion of the roster) gotten played nicely with the changes

techyCoder81 (approved):
lgtm

PR #451, opened 2022-04-06
Game and Watch changes

### LINKED ISSUE #398: [Enhancement] Make Game & Watch's Up Taunt play the ledge attack sound
Currently both up and side taunts make the same sound, but PM did a neat thing where they simply took the higher pitched bell jingle from ledge attack and made it play on utaunt, this would be a nice simple addition to have back.

### LINKED ISSUE #437: Tentative Changes for Game and Watch
Some changes I have tested that I believe can tighten Gnw's gameplay. I would really appreciate any feedback, suggestions, or tips.

- [x] Jab
	- Transition from jab 1 into jab finisher changed from pressing special to pressing attack
	- Holding attack after jab 1 will still continue the rapid jab

- [x] Up Air
	- Removed single-hit variant
	- KBG (hit 2): 95 -> 82
- Up air's KBG has been toned down as it was extremely strong for vertical confirms

- [x] Chef
	- Raised travel arc of food
		- speed_y: 0.5/0.6/0.5/0.6/0.75 -> 0.75/0.9/0.75/0.9/1.0
- Chef felt a bit lackluster as a projectile. The higher arc lets the food cover a bit more distance and allows for new pressure applications, especially on platforms

- [x] Judge
	- [x] Judge 1
		- Added air-only hitboxes
			- Angle: 270
			- BKB: 50
			- KBG: 10
		- Self-damage: 12 -> 9
	- [x] Judge 2
		- Effect: normal -> curse
		- SFX: shell -> fire
	- [x] Judge 3
		- Effect: normal -> dark
	- [x] Judge 4
		- Can be jump canceled on hit
		- Angle: 40 -> 69
		- Hitlag Multiplier: 1.0 ->2.8
		-  SFX: cut -> heavy
	- The added hitlag should make it barely reactable for confirms
	- [x] Judge 5
		- Angle (multihits): 75 -> 367
		- BKB (launcher): 40 -> 60
	- [x] Judge 9
		- Damage: 32 -> 25
		- May be a bit over-adjusted, still reliably kills around 35-40%
- The above changes are made with the goal of keeping Judge as a medium-risk medium-reward option while giving the lesser rolls a bit of utility

- [x] Up Throw
	- BKB: 100 -> 110
	- KBG: 29 -> 40
	- Endlag: 40 -> 48
- Up throw is meant to be more for positioning now that Gnw has three very strong throw mixups. It remains solid for combos at lower percents and against fastfallers, eventually transitioning to a kill throw on midweights at around 190% with optimal DI

- [x] Down Throw:
	- Damage: 4 -> 9
	- Angle: 80 -> 270
	- Effect: normal -> lay
	- Slowed down last part of animation
	- Endlag: 49 -> 43
- Down throw is now functionally identical to Snake's down throw with the difference of dropping the opponent behind Gnw and turning him around to face them after the throw. Frame advantage scales with hitstun similarly to Snake, being -5 at 0% and +6 at 161%. As Gnw's fastest option in HDR is frame 4, he has less of an advantage in the way of true confirms, but his other throws more than make up for this slight comparative disadvantage (and opponents still die at 161% from dtilt)
- The later part of the throw when the opponent's stock icon lies on the ground has been slowed down to allow opponents a bit more time to react and prepare their option
- Note: Down throw will drop opponents off ledges if grabbing very close to one, which makes them actionable immediately. Gnw is -7 in this situation, which may not be much of an issue
- For labbing purposes, down throw is +4 starting at 127% (guaranteed jab) and +5 at 144% (guaranteed up special)
- If there is a way to find the code behind the "FIGHTER_GANON_STATUS_WORK_ID_FLAG_EXPLOSION_SET_FALL" constant, this could be a lot closer to PM down throw

- [x] Up Taunt
	- Now plays the ledge attack beep to differentiate from the side taunt beep 

\
Future Tweaks
- Experiment with parachute nair or alter angles/kb
- Enlarge/lengthen lower up tilt hitboxes to be able to hit opponents out of down throw
- Add Joker's Snap and Pac-Man's Bell SFX to nus3audio to replace Judge 2 and 5 respectively

SkewedAskew:
Please see issue #437 for more information
Resolves #437
Also resolves #398


MadVillainLordQ:
Game and Watch Changes Notes

- What is the rationale behind the jab changes?
- Chef is fine overall and I'm a fan of the arc increase. I think if it does become problematic, I'd reduce bacon damage to 4% or reduce the amount of hitlag bacon induces to make comboing off of it more difficult and make it a more string oriented tool.
- Big fan of the changes on 4 and 5. Spike hammer 1 could be potentially problematic, but from my labbing it is really weak and GnW is probably negative on hit until higher percents. I do think 9 hammer should get nerf regardless of the changes made as it killing nearly at 0 is egregious. I'd advise looking towards PM values and lowering its damage first to 25% and then messing with kb curves and whatnot. 
- For uthrow, it is a little concerning how well it can kill of the top, but on correct DI (DI Away), following up with uair pass 40% is fairly difficult. I think it should be fine. 
- Dthrow looks good to me. Niche throw option but can catch people sleeping.
- One of the uair versions should be removed. Alternatively, Change the double hit uair to have slower startup, and change the animation brawl uair as having more hitlag doesn't really help distinguish between the two as the visuals are exactly the same.

Discussion was held here:
https://discordapp.com/channels/659964948365049887/950979417004466196/961371699847262299

Outcomes of discussion included potentially changing how held vs. non held inputs function. In GnW case, there are two ideas of switching nair to send at a more vertical angle while retaining strong hit uair, or changing nair to a similar move to PM GnW parachute nair with high kill power and uair retaining the juggle properties instead of the ko potential. It has not been decided yet which route to go and devs must be consulted.

techyCoder81 (approved):
looks overall reasonable to me

PR #452, opened 2022-04-06
Falcon nair readjustments

### LINKED ISSUE #449: Falcon nair retuning
Re-motion rating the move so it makes a tad more visual sense and lines up with the effects properly alongside readjusting hitbox timings/positions as necessary

black-calculus:
Readjusted the motion rates throughout Captain Falcon's nair to better mimic Melee/PM's animation which current nair is based off of, as well as retimed, repositioned, and rotated some of the effects to better suit the animation and visually represent the hitbox timing. A demonstration video is linked below:

https://youtu.be/0Ne7SoYl9lE

[Rukai data](https://rukaidata.com/P+/Captain%20Falcon/subactions/AttackAirN.html) and the [SSBWiki](https://www.ssbwiki.com/Captain_Falcon_(SSBM)/Neutral_aerial#/media/File:CFalcon_Neutral_Aerial.gif) and [Melee Frame Data](http://meleeframedata.com/captain_falcon) visualizations were referenced.

Resolves #449

### DIFF
--- a/fighters/captain/src/acmd/aerials.rs
+++ b/fighters/captain/src/acmd/aerials.rs
@@ -8,17 +8,18 @@ unsafe fn captain_attack_air_n_effect(fighter: &mut L2CAgentBase) {
     let lua_state = fighter.lua_state_agent;
     frame(lua_state, 6.0);
     if is_excute(fighter) {
-        EFFECT_FOLLOW_ALPHA(fighter, Hash40::new("sys_attack_arc"), Hash40::new("top"), 0.0, 4.5, 2.0, -148, -141, -3.0, 1.1, true, 0.7);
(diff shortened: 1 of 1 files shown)


jobrien97 (approved):
lgtm, I like the new motion rate calculation scheme

techyCoder81 (changes requested):
Seems fine to me, however with this nair design we really should get rid of the 367 angle now that the hitboxes are the size of the sun and are timed appropriately. Having this drag up/forward/down for free without respect for having to control your momentum makes this move even stronger in neutral than is probably going to be healthy. This move is historically a cornerstone of this character's boxing potential, and to have the first hit also autolink means that he can often just blaze around with it, and if he clips someone, they get dragged into him. Removing the autolink, as it has historically been, would be a healthier change overall as it would require more precision and control than it does right now.

PR #454, opened 2022-04-06
Hitlag, hit effect size, and shieldstun readjustments

### LINKED ISSUE #447: Global hitlag readjustment
Increasing base hitlag from 3F -> 4F to give moves a bit more impact and weight as well as to increase feasibility of successful DI upon getting hit

### LINKED ISSUE #448: Global shieldstun reduction
Reduce shieldstun to more reasonable levels, namely decreasing the global shieldstun multiplier from 0.9 down into the range of 0.5-0.6

### LINKED ISSUE #453: Hit effect size reduction
Reduce the size of hit effects via param modification in battle_object.prc

I'm playing a fighting game not watching an anime

black-calculus:
Specific changes:

- Global shieldstun modifier reduced 0.9 -> 0.55

- Base hitlag increased 3F -> 4F
- Animation speed multiplier during hitlag reduced 0.25 -> 0.01

- Non-elemental hit effect size scaling per damage reduced 0.037 -> 0.025
- Non-elemental hit effect base size multiplier reduced 0.52 -> 0.32
- Non-elemental hit effect minimum size multiplier reduced 0.5 -> 0.3
- Non-elemental hit effect maximum size multiplier reduced 1.1 -> 0.625
- Elemental hit effect size scaling per damage reduced 0.04 -> 0.025
- Elemental hit effect base size multiplier reduced 0.45 -> 0.37
- Elemental hit effect minimum size multiplier reduced 0.5 -> 0.4
- Elemental hit effect maximum size multiplier reduced 1.3 -> 0.875

Resolves #447
Resolves #448 
Resolves #453


jobrien97 (approved):
lgtm, added param entries are in correct order

jobrien97 (changes requested):
I'd like for hitlag adjustment to be separated from this PR, and for it to be its own branch.

I feel base hitlag increase should be kept in a draft PR until we can re-tune all of the per-move hitlag modifiers to feel good with the increase base hitlag

techyCoder81 (approved):
I disagree, I would rather have this change in, and be able to play with it in nightlies and such, and improve the hitlag over time. If the hitlag changes get "kept in a separate branch until we have time", we simply won't end up actually making the time, and it will rot in the corner.

I am also wondering though, what was our shieldstun value originally, 5 months ago before we increased it? 0.9 -> 0.55 seems like a pretty substantial change, and I would expect something in the middle would be more appropriate.

JoaoAlmeida20:
I like the hitlag, only have some doubts on multi-hits, where a lot of them start to feel slightly syrupy. But I feel like for some of them that's fine, it's not comparable to vannila hitlag where multi-hits become cutscenes, and for ones where it stays a problem (maybe even balance wise because of becoming too trivial to SDI out of them) they can get some personal adjustments eventually
The hit effects are great
Don't have much an opinion on the shieldstun from just fighting CPUs, but I agree with techy that 0.9 to 0.55 seems like a really  massive leap

PR #457, opened 2022-04-06
Banjo changes

black-calculus:
A few QOL changes for Banjo for the next beta. Changelog attached.
[Banjo Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8423969/Banjo.Changelog.txt)


fighters/buddy/src/opff.rs
```diff
@@ -23,10 +23,42 @@ unsafe fn grenade_ac(fighter: &mut L2CFighterCommon) {
     }
 }
 
+// Banjo Dair bounce
+unsafe fn dair_bounce(fighter: &mut L2CFighterCommon, boma: &mut BattleObjectModuleAccessor, motion_kind: u64, frame: f32){
```
jobrien97: Should really only be importing L2CFighterCommon now, and using fighter exts accordingly

fighters/buddy/src/opff.rs
```diff
@@ -23,10 +23,42 @@ unsafe fn grenade_ac(fighter: &mut L2CFighterCommon) {
...
+    && frame < 45.0
+    {
+        if AttackModule::is_infliction_status(fighter.module_accessor, *COLLISION_KIND_MASK_HIT | *COLLISION_KIND_MASK_SHIELD) {
+            MotionModule::set_frame_sync_anim_cmd(boma, 45.0, true, true, false);
+        }
+    }
+}
+
+// Banjo Wondering Fail on command
+unsafe fn wonderwing_fail(fighter: &mut L2CFighterCommon, boma: &mut BattleObjectModuleAccessor, situation_kind: i32, frame: f32){
```
jobrien97: Same here, unused imports too

fighters/buddy/src/opff.rs
```diff
@@ -23,10 +23,42 @@ unsafe fn grenade_ac(fighter: &mut L2CFighterCommon) {
...
+    wonderwing_fail(fighter, boma, situation_kind, frame);
+}
+
 #[utils::macros::opff(FIGHTER_KIND_BUDDY)]
 pub unsafe fn buddy_frame_wrapper(fighter: &mut L2CFighterCommon) {
     common::opff::fighter_common_opff(fighter);
-
+    buddy_frame(fighter);
     blue_eggs_land_cancels(fighter);
     grenade_ac(fighter);
```
jobrien97: Should move these to moveset func

fighters/buddy/src/acmd/aerials.rs
```diff
@@ -118,10 +115,100 @@ unsafe fn buddy_attack_air_b_game(fighter: &mut L2CAgentBase) {
...
+        ATTACK(fighter, 3, 0, Hash40::new("k_wingl3"), 7.0, 76, 90, 0, 50, 4.5, 6.5, -3.0, -1.0, None, None, None, 1.0, 1.0, *ATTACK_SETOFF_KIND_ON, *ATTACK_LR_CHECK_POS, false, 0, 0.0, 0, false, false, false, false, true, *COLLISION_SITUATION_MASK_GA, *COLLISION_CATEGORY_MASK_ALL, *COLLISION_PART_MASK_ALL, false, Hash40::new("collision_attr_normal"), *ATTACK_SOUND_LEVEL_M, *COLLISION_SOUND_ATTR_BUDDY_WING, *ATTACK_REGION_PUNCH);
+        ATTACK(fighter, 4, 0, Hash40::new("top"), 7.0, 76, 90, 0, 50, 3.75, 0.0, 11.0, 0.0, None, None, None, 1.0, 1.0, *ATTACK_SETOFF_KIND_ON, *ATTACK_LR_CHECK_POS, false, 0, 0.0, 0, false, false, false, false, true, *COLLISION_SITUATION_MASK_GA, *COLLISION_CATEGORY_MASK_ALL, *COLLISION_PART_MASK_ALL, false, Hash40::new("collision_attr_normal"), *ATTACK_SOUND_LEVEL_M, *COLLISION_SOUND_ATTR_BUDDY_WING, *ATTACK_REGION_PUNCH);
+    }
+    frame(lua_state, 9.0);
+    if is_excute(fighter) {
+        
+    }
```
jobrien97: ?

MadVillainLordQ:
Banjo Notes

- Double Jump change feels very minimal. Maybe consider allowing banjo to use his third jump sooner in his double jump?
- Dair needs to be more punishable on shield. Currently it is -1. You should also reduce trip chance for this move to 0. It is +24 on hit at 0 in the air and +35 at 50. My suggestion is to add 5 more frames of endlag to when banjo can act in the air.
- Uair looks fine. Personally prefer it potentially have some ko power but if you prefer to solely juggle/setup into fair its serviceable for it. 
- Fair changes are good as well. I do like the distinction with sweetspot/sourspot fair but I think it would work better/give it more flavor if sourspot had less kb (kinda like daisy sour/sweet fair). Maybe some might miss removal of meteor, but I'm indifferent about it.
- Fthrow/Bthrow/Dthrow (regular) are fine as well
- Dthrow (bury) - This is my own biased opinion as I am not a fan of buries in general but I think they're especially badly designed when attached to throws. I think a unique idea you could do with this bury if you decide to keep it is to have it waste a feather of wonderwing for a "guaranteed" punish. Would be really cool resource management that way, and the animation you have for it visually shows that a feather is being used. Although if asked my preference, I'd straight up remove the bury and keep the new dthrow you designed.
- For wonderwing, can you explain why you increased the kb of the move? It was already really potent and has a ton of other utility attached to it. I know its attached to feathers but it really does do a lot and I don't think it needed more KO power. Can you also explain the rationale for adding a trip? Don't know if that's a reference to his game, but a fake out like that is pretty messed up especially near ledge since he can fake that out when coming back to stage.

Discussion with Calc on these points:
https://discordapp.com/channels/659964948365049887/950979417004466196/961411365220810762

Potential change to down throw bury is have it consume a feather.

techyCoder81:
agree on bury throws being inherently unhealthy in this engine, this line is concerning:
`~ [!] Holding the attack button before initiating the throw now activates a slower throw that buries`
That said, making it attached to a feather would be more reasonable if its actually feasible, since then you won't have to worry about stuff like ROB's dthrow shenanigans.

fighters/buddy/src/opff.rs
techyCoder81: Agreed, for characters that we touch who have *really* simple OPFF files, we should clean up the 1-2 existing calls that they have and get rid of superfluous "frameinfo" bs and "moveset" functions when possible.

JoaoAlmeida20:
- I disagree on up air, I feel like this animation it has has always looked super weak so having more kill power would feel off, and it doesn't seem necessary to me. I'd honestly consider even going a bit in the opposite direction and reducing the kbg a bit to make it juggle for longer, leaning a bit more into something like Roy's up air territory, specially since his double jumps are still so weak he quickly starts to struggle to keep up with the opponent after an up air at later percents.
- I also dislike bury throws so I agree it may be unnecessary specially because the new dthrow is pretty solid.
- About the funny tripping feature, I really like it, but it seems like in the air, offstage, its a bit too useful as a juke, because you can grab ledge pretty fast and don't lose that much height during the process either. On the ground, I feel like something weird is up, from what I remember in vanilla you do have access to get up attack but here I was unable do anything. Banjo just got up quickly, I couldn't do getup attack nor roll. Not as funny

jobrien97:
Most recent commit is misnamed - should be "change Wonderwing fakeout to shield input/add eggs firing landing smoke gfx"

black-calculus:
Made some final updates for now, here are the changes from the previous PR:
[Banjo Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8674437/Banjo.Changelog.txt)

And the full final changelog for this set of changes:
[Banjo Changelog (Full).txt](https://github.com/HDR-Development/HewDraw-Remix/files/8674439/Banjo.Changelog.Full.txt)

jobrien97 (approved):

jobrien97:
As of now, you can shield in between Banjo's jab combo

jobrien97:
As of now, you can shield in between Banjo's jab combo

PR #462, opened 2022-04-07
Reinstate vanilla cstick IRAR behavior

### LINKED ISSUE #458: IRAR performed with C-stick is no longer functional
In the previous patch you were able to Dash and perform an IRAR if you flicked your c-stick in the opposite direction and jumped on the same frame. That is no longer possible in this patch. Calc has mentioned an issue with JOB's code that may have been the cause of this. 
https://discordapp.com/channels/659964948365049887/950979364634382376/961022673855602758

jobrien97:
During dash, inputting cstick backwards then jump within next 2 frames will allow a full momentum IRAR, as in vanilla.

Resolves #458


MadVillainLordQ:
Tested in game and works like a charm

techyCoder81 (approved):
heavily discussed in discord, approved with the caveat that it could always get deprecated later at some point

PR #464, opened 2022-04-07
Fix tethers not being edgehoggable

### LINKED ISSUE #439: Tether ledgetrump works the opposite way it's supposed to
When I ledgehog characters with tethers in the new beta, I get ledgetrumped instead of them

black-calculus:
Resolves #439 

Also readjusts the params on trump momentum and actionability to suit the more consistent behavior on when you get trumped, sending you less high and adding 5 frames of endlag.


MadVillainLordQ:
Fix works in almost every scenario except when tethered person grabs a frame on/after an opponent grabs ledge. The opponent on ledge is ledge trumped and placed in trump lag. 

Possible explanation for trumping on the same frame is port priority but I could not test to make sure. 

See discussion here:
https://discordapp.com/channels/659964948365049887/950979417004466196/961851691433594890

jobrien97 (dismissed):
lgtm

techyCoder81:
Does this fix appropriately handle ICs?

techyCoder81:
looks like this fix once again does not correctly take ICs into account. I'd like to see this character's edgehog behavior finally work tbh. It hasn't work correctly for 3 years now.

techyCoder81 (changes requested):
Looks like this change does not correctly take ICs into account. This character's ledgehog has not functioned for 3 years and I would like to see it finally work at some point.

techyCoder81:
you can get the partner boma by using `object.battle_object_id + 0x10000`, such as 
```
let partner_id = (popo_boma.object().battle_object_id + 0x10000);
let partner_object = unsafe { &mut *get_battle_object_from_id(partner_id) };
```

jobrien97 (approved):
ya looks great FINALLY

techyCoder81 (approved):

PR #467, opened 2022-04-09
Mythra up-special cancel

blu-dev:
This should not be merged until the angles have been put into params and the cancel frame is set in a param file

This changes mythra's recovery to have have default behavior when neither A or B have been triggered/press. The default behavior is that she retains her jump momentum into an actionable fall, with side-b and up-b disabled again until she either grabs a cliff or a ledge.

It's entirely possible that this change is b r o k e n in which case there are ways to make it less so. Should also be noted that her up special can now be angled 30 degrees forward and 10 degrees back, and that the frame 1 hitbox has been moved to frame 2 in order to follow mythra's direction better. The fixed knockback has been increased on the first hit only, the rest of the hitboxes have only been pushed back a frame.

This is linked to HDR-Development/hdr-private#68


fighters/elight/src/status/special_hi.rs
```diff
@@ -0,0 +1,116 @@
+use super::*;
+
+pub unsafe fn special_hi_common_check_spreadbullet(fighter: &mut L2CFighterCommon) {
+    // [v] this is likely a leniency check so that if you accidentally double tap
+    //      the special button you aren't forced into spreadbullet. it should be
+    //      noted how here it requires the `trigger` instead of holding it, so if you
+    //      hold the button it also doesn't trigger until right before the move would come out
+    let frame = dbg!(fighter.get_int(*FIGHTER_ELIGHT_STATUS_SPECIAL_HI_INT_FRAME_FROM_START));
```
techyCoder81: should this dbg be here?

techyCoder81 (dismissed):
very, very cool overall

techyCoder81 (approved):

PR #482, opened 2022-04-13
Sephiroth changes

black-calculus:
A few QOL changes for Sephiroth for the next beta, bringing his hitboxes and active frames to a much more reasonable level while giving some long-awaited buffs to a few other parts of his kit. Changelog attached.
[Sephiroth Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8481776/Sephiroth.Changelog.txt)


techyCoder81 (dismissed):

JoaoAlmeida20:
The jab change is awesome but the hitbox added partway through the blade seems to extend a bit too much into the backside or something as it can cause wonky reverse hits and sometimes scoop opponents behind you
Also about his dtilt, would you consider taking a look at it's late hit? There seems to be an agreement that a lot of these sliding dtilts are a little overtuned in hdr and imo it's because of the late hit being as big as the early hit, it makes the move too good at challenging ledge with 0 timing involved and challenging the move in general feels too hard. 
I made an issue about it earlier so that each character could be seen case by case if any change is actually called for: https://github.com/HDR-Development/HewDraw-Remix/issues/483

techyCoder81 (approved):

PR #485, opened 2022-04-16
Samus glows blue when shinespark is ready

JoaoAlmeida20:
Samus glows blue after running for 30 frames to indicate that shinespark is available


techyCoder81 (approved):
looks good to me

PR #492, opened 2022-04-18
Mk rework and Meta Quick mechanic

techyCoder81:
various adjustments to MK frame data and hitboxes
side special now triggers immediately on hit, and is actionable as a combo starter
New mechanic: meta quick! 
- After dealing 50 damage, you will gain Meta Quick as an install which can be triggered with taunt (while not in shield).
- Upon inputting taunt, you gain generally increased movement speed, travel distance, and acceleration across your whole kit.
- for example, side special, up special, dash, run, air accel, and roll all go farther/faster when in MQ
- This install lasts for 8 seconds
- dealing additional damage extends the length of meta quick (0.5s per 5 damage dealt)
- Once stacked, meta quick remains available after death, but will be lost if you die while it has been activated.


SkewedAskew:
Notes from initial testing:
- Meta Quick state and current damage dealt persists through death as well as salty runback, flash also persists on victory screen
- Meta Quick does not apply increased air accel on grounded jump, which feels fine if intended
- Accel is being (mistakenly?) applied to Drill Run, this version does not ever connect all its hits and travels extremely far
- MQ Shuttle Loop has insane ladder potential off stray hits such as uair, dtilt, usmash with either first or second hit. Would recommend further kb/angle adjustments. Another suggestion is making MQ version shorter than normal as compensation for the increased horizontal mobility
- Can wall of pain pretty easily with the increased accel
- A possible suggestion would be to make attacks deal reduced damage during MQ, making moves less safe and therefore lessening its threat when not already in advantage state. Another option could be worsening non-MQ combo potential, making finisher conversions weaker or less reliable without it
- Awesome job on initial implementation, feels very fun

JoaoAlmeida20:
Initial impressions

- The 'fight me' sound + the stop all sounds doesn't seem to be working very nicely;
  With up taunt, there is this weird mumbling sound and nothing else
  WIth side taunt, he says both 'fight me' and 'come' a bit overlapped
  With down taunt, there are only sword swoosh sounds

- Momentum transfer on grounded jump during meta quick doesn't seem to be working, in fact it seems it's even slower than his meta quick air speed. Maybe he's getting capped to his regular not meta quick jump speed, which might be lower than his meta quick air speed

- Up b seems a bit wild, I'm not sure how it was before but the way it can drift around so much seems to make it trivial to confirm from most hits, but I didn't test with another player so maybe it's necessary for it to even be a relevant move with DI and such.

- About the available indication flash, I feel like it could be a bit more transparent, and comparing to Wario's waft vfx, I feel like the first time Wario's waft is ready the first flash has some extra oomph to it? There may be some sort of extra flash, similar to the one used for the gimmick flash at the moment. Could be good maybe

SkewedAskew:
- Non-MQ momentum on dash/run/airspeed/Drill Run now seems to vary. Jumps three and five in particular have noticeably more momentum than the others. I found consistent instances of extra slow speed being applied by running immediately after a smash attack
- Non-MQ Shuttle Loop feels a bit too weak now, DI away has midweights living until past 150% even when connecting the second hit near the top blastzone

SkewedAskew:
Meta Quick still persists through death and on the win/no contest screens, but if you fix that then I'd say this is ready for merge. Feels very smooth now, great work!

techyCoder81:
Good catch I forgot about that

SkewedAskew:
- MQ charge persists through salty runback
- Consider lowering the hitlag of Drill Rush proc from 3.0 to the 1.5-2.0 range, it feels a bit jarring

Other than that, all good

fighters/common/src/function_hooks/dash_dancing.rs
```diff
@@ -179,7 +179,16 @@ pub unsafe fn status_dash_main_common_hook(fighter: &mut L2CFighterCommon, arg:
     let run_accel_mul = WorkModule::get_param_float(boma, hash40("run_accel_mul"), 0);
     let run_accel_add = WorkModule::get_param_float(boma, hash40("run_accel_add"), 0);
     let ground_brake = WorkModule::get_param_float(boma, hash40("ground_brake"), 0);
-    let dash_speed: f32 = WorkModule::get_param_float(boma, hash40("dash_speed"), 0);
+
```
jobrien97: This file is unused, should honestly be deleted

jobrien97:
I'm questioning the use of VarModule here to store speed multipliers.
Vanilla's install speed modifiers work by editing a character's movement params directly within WorkModule::get_param_*
Would it not be cleaner to use the get_param hooks to modify MK's movement params while is_meta_quick() returns true? Applying multipliers to his control/motion energies seems messy and could have potential edge cases.

fighters/metaknight/src/meta_quick.rs
```diff
@@ -0,0 +1,205 @@
...
+        VarModule::set_float(fighter.object(), vars::common::JUMP_SPEED_MAX_MUL, 0.75);
+
+        // if you are initial dash, slow them down slightly
+        if fighter.is_status(*FIGHTER_STATUS_KIND_DASH) {
+            let motion_vec = Vector3f {
+                x: -0.25 * PostureModule::lr(fighter.boma()) * (1.0 - (MotionModule::frame(fighter.boma()) / MotionModule::end_frame(fighter.boma()))),
+                y: 0.0, 
+                z: 0.0
+            };
+            KineticModule::add_speed_outside(fighter.boma(), *KINETIC_OUTSIDE_ENERGY_TYPE_WIND_NO_ADDITION, &motion_vec);
```
jobrien97, line 63: pls don't use wind anymore lol, use sv_kinetic_energy funcs (add_speed in this case)

techyCoder81:
need to remove printlns and the unused dash mul const

fighters/metaknight/src/meta_quick.rs
techyCoder81: i used this for a reason but i can no longer remember what it is so I guess ill just look at it again

fighters/common/src/misc.rs
```diff
@@ -26,43 +26,50 @@ pub fn fighter_reset(fighter: &mut L2CFighterCommon) {
...
+    let fighter_kind = fighter.global_table[FIGHTER_KIND].get_i32();
+    let jump_speed_x = WorkModule::get_param_float(fighter.module_accessor, hash40("jump_speed_x"), 0);
+    let jump_speed_x_mul = WorkModule::get_param_float(fighter.module_accessor, hash40("jump_speed_x_mul"), 0);
+    let run_speed_max = WorkModule::get_param_float(fighter.module_accessor, hash40("run_speed_max"), 0);
+    let stick_x = ControlModule::get_stick_x(fighter.module_accessor);
+    let ratio = VarModule::get_float(fighter.battle_object, vars::common::JUMP_SPEED_RATIO);
+    //println!("run_speed_max: {}", run_speed_max);
+    //println!("jump_speed_ratio: {}", ratio); 
+
+    // get the multiplier for any special mechanics that require additional jump speed max (meta quick, etc)
```
techyCoder81, line 38: except for these few lines here, this is just a tab formatting fix

fighters/metaknight/src/meta_quick.rs
```diff
@@ -0,0 +1,221 @@
+// opff import
```
techyCoder81, line 1: this file is the meat of the logic for MQ.

scripts/build.py
```diff
@@ -26,14 +26,14 @@ def install_with_ip(ip: str, is_dev: bool):
     if not is_dev:
       os.system('curl.exe -T ..\\plugin\\target\\aarch64-skyline-switch\\release\\libhdr.nro ftp://' + ip + ':5000/ultimate/mods/hdr-dev/plugin.nro')
     else:
-      os.system('curl.exe -T ..\\plugin\\target\\standalone\\aarch64-skyline-switch\\release\\libhdr.nro ftp://' + ip + ':5000/ultimate/mods/hdr-dev/plugin.nro')
```
techyCoder81, line 29: this is reordered so that the dev nro gets sent first (so you dont have to wait 15 seconds waiting for main plugin to get curled first lol

dynamic/src/ext.rs
```diff
@@ -680,6 +686,17 @@ impl BomaExt for BattleObjectModuleAccessor {
         ModelModule::set_joint_rotate(self, Hash40::new(&bone_name), &rotation, MotionNodeRotateCompose{_address: *MOTION_NODE_ROTATE_COMPOSE_AFTER as u8}, MotionNodeRotateOrder{_address: *MOTION_NODE_ROTATE_ORDER_XYZ as u8})
     }
 
+
+    /// gets the FighterKineticEnergyMotion object
+    unsafe fn get_motion_energy(&mut self) -> &mut FighterKineticEnergyMotion {
```
blu-dev, line 691: these should probably be changed in skyline-smash but it's fine here for now

fighters/common/src/function_hooks/dash_dancing.rs
```diff
@@ -1,607 +0,0 @@
-use super::*;
```
techyCoder81, line 1: this file is not installed and is unnecessary with our new dash implementation

blu-dev (approved):
lgtm, didn't check ACMD because beeg diffs but I assume they are fine

Riah321:
Are the changes at the top of this page all the changes? If not please make a proper changelog for the next beta :)

PR #498, opened 2022-04-20
Palu Reflecter Training Mode Fix

### LINKED ISSUE #479: Have Palutena aegis timer reset on training mode restart
Along with the death check flag, add `smash::app::sv_information::is_ready_go() == false` as a check, so it resets on training mode restart, which feels more dynamic than having to taunt first.

Wish325:
Along with the death check flag, add `smash::app::sv_information::is_ready_go() == false` as a check, so it resets on training mode restart, which feels more dynamic than having to taunt first.

Fixes #479


techyCoder81 (approved):
lgtm

PR #507, opened 2022-04-24
Fix crawl momentum from right to left and when you have backward speed

### LINKED ISSUE #506: Crawl maintains momentum differently depending on if the character is going left to right or right to left
Crawl is able to maintain momentum, but from right to left it does not work and loses it almost instantly

JoaoAlmeida20:
Rewrites the squat_walk init function so that it handles the calculation of the crawl animation speed multiplier properly.
Edit: To elaborate, handling the calculation properly means that you maintain the speed you had and can reverse it in either direction, instead of only being able to do that from left to right. Additionally, you can also maintain backward speed.
Resolves #506

### DIFF
--- a/fighters/common/src/general_statuses/crawl.rs
+++ b/fighters/common/src/general_statuses/crawl.rs
@@ -1,0 +1,138 @@
+// status imports
+use super::*;
+use globals::*;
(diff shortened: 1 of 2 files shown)


jobrien97 (changes requested):
remove printlns pls

jobrien97 (approved):
lgtm

PR #509, opened 2022-04-26
Knockback gravity and tumble readjustments

### LINKED ISSUE #469: Knockback behavior readjustments
- Adjust params for gravity-based additional launch speed as well as combo gravity/fallspeed for all characters as appropriate
 -  Increase tumble threshold so low percent knockback isn't sending into tumble quite as early and DI/active defensive options (jump, attack, airdodge, etc.) are more feasible escape tools at low-mid percent to help mitigate the rampant early tech situation problem

black-calculus:
Increased the tumble threshold from 80 units of KB to 90 and set a lower and upper cap on the hitstun gravity for all characters at 0.09 and 0.15 respectively to make comboing super floaties and super fastfallers less obnoxious and to make being comboed as them less frustrating. Additionally increased the added gravity-based vertical launch speed that applies during angles that `damage_fly_top` does not take override fall speed and gravity, which is currently angles of 45-135.

Resolves #469


techyCoder81 (approved):

techyCoder81:
heard positive comments about how it feels in discord

techyCoder81:
ticket is not in draft state, its been play tested, has an approval, so I am merging it

PR #511, opened 2022-04-26
Link adjustments

techyCoder81:
- reduces the size of both grounded and aerial up special vfx and improves their placements and timings
- nair active frames reduced 29 -> 18 (comparable to fox)
- fair has slightly more time in between the two hits (this also improves optimal sh fair1 fastfall fair2 timings)
- fair hit 2 has increased damage, kb, and shield damage, though still not quite as much as fair1
- fair hit 1 and hit 2 hitboxes and animation timings improved to better match the vfx (new sword trail is in the next nightly to coincide with this)
- boomerang has slightly increased bkb
- grounded uspecial shieldstun is reduced
- grounded uspecial late hits have had their angles increased 20 -> 30, 1 -> 15
- dsmash back hit angle reduced to 30 degrees behind link and knockback rescaled appropriately (reference kb was smash 4 back hit)
- dsmash front hit bkb increased
- dsmash endlag reduced
- utilt now interpolates properly
- utilt size reduced to match sword length
- dtilt active frames reduced 5 -> 3
- bair both hits will always send backwards (it cannot reverse hit)
- bair lasts longer and has slightly more time between the hits
- dash attack sourspot damage (and kb) increased
- dair endlag reduced and auto cancel window moved earlier, can now auto cancel out of jump -> dj dair (for reference)
- ftilt no longer hits behind link, but it also no longer has a large blind spot 45 degrees in front of him (where there is a prominent sword trail)
- various effects and sword trail improvements


ChaserTechHDR:
Neutral Air feels like the animation should match with it's active frames, similar to Melee but without the active frames working when Link retracts his foot. So either the animation needs to be shorten or the hitbox should revert back to normal.

And as for Ftilt, I personally prefer the the move hitting behind Link like it used to as well. The other changes seem pretty reasonable though.

PR #523, opened 2022-04-29
Sephiroth Jab Recovery 25 > 12

CodenameComlink:
should reduce Sephiroth's Jab recovery by 13 frames


techyCoder81:
closed per request of comlink

PR #541, opened 2022-05-03
Sora

### LINKED ISSUE #540: Sora rework
Reworking Sora for the next beta, as discussed in length with Wish. More details can be found on the PR linked to the issue.

black-calculus:
Resolves #540 .
Big batch of Sora changes for the next beta. Got one last final coming up and a lot more characters to go through and put up PRs for so I kinda rushed through this changelog, hopefully it's not missing anything.
[Sora Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8608384/Sora.Changelog.txt)

You can grab the animations and motion_list you need from the [latest nightly romfs release](https://github.com/HDR-Development/romfs-release/releases/tag/v0.16.0).


fighters/trail/src/opff.rs
```diff
@@ -34,23 +86,54 @@ unsafe fn fair_cancels(boma: &mut BattleObjectModuleAccessor, cat1: i32, status_
...
+        }
+    }
+}
+
+// Actionability after hitting aerial sweep
+unsafe fn aerial_sweep_hit_actionability(boma: &mut BattleObjectModuleAccessor) {
+    if boma.is_status(*FIGHTER_STATUS_KIND_SPECIAL_HI){
+        if MotionModule::frame(boma) > 37.0 {
+            if AttackModule::is_infliction(boma, *COLLISION_KIND_MASK_HIT) {
+                DamageModule::add_damage(boma, 1.0, 0);
```
jobrien97: Why are we adding damage here?

jobrien97 (changes requested):
remove printlns pls

MadVillainLordQ:
Overall:
Really well done from a design standpoint. The following are notes/suggestions to his moveset

Notes:
- At some point, Firaga should be changed to not halt vertical momentum. It is extremely strong to be able to shoot a projectile at ground level and waveland immediately afterwards
- Add reverse hit properties on the new dair, specifically on frames 8-9 as this style of hitbox is a catch all move and removes the potential nuance of DI when hitting an opponent. 
- The suggestion would be to make frames 8-9 reverse, 10-12 hit normally, and frame 13 be the sole spike hitbox and only place the spike on the blade of kingdom key, not on the hilt or body of sora. Another idea is to have only frame 8 be the reverse, 9-12 regular, and 13 be spike (with the aforementioned hitbox placement)
- It has been mentioned in discord channels but buffering short hop neutral b with thunder causes grounded thunder to appear instead of aerial thunder
- I have concerns on the potential looping of fair 1/2 as they are +22 and +15 on hit at 0. Hitstun should be reduced on those moves to avoid that.
- Firaga hitstun is a tad high at 34 frames. Being able to waveland out of it makes it have much more frame advantage than its current form. It should be brought down to be +24 to be more inline with moves that have similar properties like wolf laser (+22)
- Uair damage is a concern of mine as the move is already very good at comboing and is further improved with the attribute changes that sora has had. I believe it should be brought down to 10 damage to be more inline with other arc sword swipes of similar nature
- Awesome change with the proration of moves after fair 2 cancels. I do think the moves have to have even more damage reduction though. 
- It would be nice if there was some sort of visual/audio cue when you hit the spike on dair. Maybe like the little ding noise from his stop taunt could work

PR #548, opened 2022-05-06
Zelda changes

black-calculus:
Big batch of Zelda changes for the next beta, changelog can be found right here: 
[Zelda Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8638399/Zelda.Changelog.txt)

You can grab the motion_list you need from the [latest nightly romfs release](https://github.com/HDR-Development/romfs-release/releases/tag/v0.17.0).


MadVillainLordQ:
Zelda Notes

- Fair active frames from 6-17/bair active frames 8-19/ Current fair is 8-20 and bair is 8-21. I think we should consider reducing active frames of the late hitboxes of both moves a bit more. My suggested time would be for fair to have frames 6-7 strong hit/8-12 weak hit and bair to have 8-9 strong/10-18 weak. 
- Dair active frames could serve to be reduced from 14-24 to 14-21 in which frames 14-15 are strong and 16-21 are weak. I'd say this change is justified as there was a massive improvement to her faf and AC on the move and it would be fairly strong to have a spike be active for 11 frames. 
- Outer Ftilt/Fsmash kb values should be switched. (this is not to say actually switch the hitbox kb stats but have it emulate each other. This gives more of a reason to use fsmash over ftilt as ftilt currently hits harder than uncharged fsmash. I will upload a picture to demonstrate their different kbs on marth at 80 percent)
- Jab is consistently -9 on shield instead of the numbers suggested by in the changelog
- Remove vacuum hitbox property of utilt and make it send in one direction only (with current angle) or just have it send straight up at 90. (I'd suggest for it to send at 90 as ftilt already sends behind her and nayru sends in front)
- Jab vfx should be pushed out a bit further to better represent the jab hitbox (will upload picture that shows vfx vs. hitbox placement)
- Dash attack hitbox was pulled in from the last patch. I think outer hitbox should match previous outer hitbox reach and chest hitbox pushed forward by the same amount to better cover zelda's arms (will upload picture to demonstrate)
- I have a concern with how max charge phantom functions as I think being able to use phantom at any point other than the current restrictions you mentioned to be potentially very oppressive. I'd suggest to have her still do the phantom release animation but motion rate it so that the animation lasts anywhere from 15-20 frames as opposed to 40 frames it currently has. I'd even go so far to say you can do this with every charge version if you think that's more appropriate. Even a special cancel feature (cancel any move on hit/shield which is similar to nayru) with the 20 frame animation is a more healthy design
- Dins fire change is very cool. I think you can consider increasing the initial hitbox size to the low medium size of dins fire (the size that occurs after about 10 frames of dins fire being summoned) and potentially increase minimum damage to 5% and scale appropriately to a max damage at 15%.

black-calculus:
Gonna merge the PR for the time being and note all of these down for the next round of changes we do when we get to Zelda.

PR #549, opened 2022-05-06
[Bugfix/Change] Waveland Reimplementation

### LINKED ISSUE #515: Disable airdodge ground snapping after stall is over
Once kinetic type is KINETIC_TYPE_FALL

blu-dev:
1. Deprecated the old waveland check code in favor of `handle_waveland` in the `BomaExt` trait (so it can be reused elsewhere with ease.
2. Modified the waveland code to *not* loop while setting your position lower and lower, which at points in the past would cause infinite loops/hangs.
    - It now is a singular line-segment test down by the units specified by the waveland threshold in `fighter/common/hdr/fighter_param.xml`
    - It uses the bottom of the ECB rhombus instead of `PostureModule::pos`, which should be more accurate/meaningful positioning to base waveland detection on
    - Uses the actual ground collision instead of whatever the "floor" is considered

Resolves #515

### DIFF
--- a/dynamic/src/ext.rs
+++ b/dynamic/src/ext.rs
@@ -413,7 +413,7 @@ pub trait BomaExt {
     unsafe fn get_param_int64(&mut self, obj: &str, field: &str) -> u64;
 
     // tech/general subroutine
-    unsafe fn handle_waveland(&mut self, change_status: bool) -> bool;
+    unsafe fn handle_waveland(&mut self, require_airdodge: bool, change_status: bool) -> bool;
 }
 
 impl BomaExt for BattleObjectModuleAccessor {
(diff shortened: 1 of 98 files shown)


JoaoAlmeida20:
Regular airdodges seems to be working nicely as expected, there are some times with some upward airdodges where with the old implementation you'd expect to get a waveland and here you seem to not get it but they are pretty situational and it could probably be considered an improvement to not have them.
However the bowser jr up special airdodge scenario is not working very well. You sort of float there for a while before going into up b landing animation:
https://streamable.com/6t7v31
You used to be able to waveland into the up b landing animation by doing the timing shown in this video. It is to be fair a pretty weird behavior it had, considering regular neutral airdodges can't waveland and the only airdodge bowser jr gets during up special is neutral airdodge, but it makes more sense than for him to awkwardly float there for a bit

JoaoAlmeida20:
Bowser jr up b airdodge wavelanding works as expected now, going immediately into his up b landing animation as it should.
Also this implementation seems on initial testing to fix wonky wavelands/ledgedashes on some of HDR's custom stages. Tested on mushroom kingdom U which I think sometimes would have that thing where you airdodge across the ground if you do it from ledge and that did not happen

jobrien97 (approved):
lgtm, feels nice ingame

PR #555, opened 2022-05-10
Fix ECB shifts

### LINKED ISSUE #554: ECBs do not shift upwards properly for characters other than Samus and Dark Samus
Title is self-explanatory. Vector for failure has not been identified yet, most likely is in the ECB shift code in fighters/common/opff/physics.rs

black-calculus:
Dunno how Samus and Dark Samus were the only ones the old code worked on...

Anyway, this should fix ECB shifts for everyone.
Resolves #554


blu-dev (approved):
not sure why it wasn't working before but if this fixes it, it fixes it. thanks

PR #558, opened 2022-05-12
Byleth changes

black-calculus:
Big batch of Byleth changes for the time being. Will post the changelog and link to the updated romfs once those are up.


black-calculus:
Changelog attached here:
[Byleth Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8674870/Byleth.Changelog.txt)

You'll need the updated romfs as well:
https://github.com/HDR-Development/romfs-release/releases/tag/v0.17.3

SkewedAskew:
- Up special can chain into itself on floatier characters as demonstrated here (tested on all DI):
https://user-images.githubusercontent.com/55216313/168202653-f5240bf9-6ffd-454d-a740-6c6c80d2d528.mp4
Suggestion would be to alter the angle or to make some sort of flag to disable a subsequent grab.

- Additionally, I encountered an issue where connecting with the later part of the up special while an opponent is in tumble will override both followup choices and eject Byleth to the side with the opponent completely actionable. Occurs with both grounded and aerial versions

- Rest of the changes look great

Baadwulf:
Other than the up b chains mentioned above LGTM.

Riah321:
Other than the up b chains mentioned above LGTM.

SkewedAskew:
- Up special is better, but I would consider lowering the angle a little bit more. It can still chain into itself on floaties and midfallers (fastfallers can escape with DI out)
- I would also tone down the knockback of aerial light input side special. Up special confirms into tipper side special on all characters for very early stocks

PR #567, opened 2022-05-12
Steve changes

black-calculus:
Small batch of Steve changes for the time being. Will post the changelog and link to the updated romfs once those are up.


kalomaze:
Doesn't crash and seems to function properly. 
Some QoL changes that I would do in the future (they don't all have to be done for this pr):
- Soul fire's longer startup is enough of a visual indicator that I think we don't need a flame to spawn next to his hand unless that was unintentional.
- Make up air & nair fastfallable, currently both act like pacman's up air which is strange because it doesn't lift steve up at all just like his fastfallable fair and bair.
- Buffered Short hop fair still has the trail from his buffered backwards nair and should be adjusted. Maybe get rid of the mechanic entirely as well so he can buffer bair properly?
- Change block place criteria to be for whichever direction he's placing it so that he can place blocks upward or to the side even when he's barely above ground
- Held block place input should lock the original block placement direction for the remainder of the held input: tap input should be used to rapidly place blocks in different directions
- Make block place frame 3 instead of frame 6, it can be annoying to deal with the directions new block place allows without any way to buffer it consistently

kalomaze:
Nice, btw, that list of steve changes i had was a wishlist of sorts of ideas after labbing and not really stuff that im necessarily gonna campaign for to be in cause it might not even be great design in the first place. i might tackle some of them in the future. Don't wanna come across as demanding with how i worded it

black-calculus:
Changelog:
[Steve Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8748384/Steve.Changelog.txt)

PR #574, opened 2022-05-12
Bayonetta changes

black-calculus:
Big batch of Bayonetta changes for the time being. Will post the changelog and link to the updated romfs once those are up.


MidRush:
Bayonetta changes feel very fair for the most part. Up air changes are really fun without breaking the character. Additional startup on DABK makes it easier to input Side B > down instead of the Hadoken motion. Love the speed and momentum buff on it too, as if the extra startup is charging a more powerful kick. 

Up tilt feels a frame too laggy, makes the move feel a bit stiffer than I would personally prefer. Maybe could just have endlag reduced by one frame imo. 

So far no crashes.

EDIT: Downwards hitbox might be overtuned. Feels like it can pick up combos from further below me than she should be able too. Only suggestion would be maybe moving the hitbox from the foot to the thigh. There should be incentive to fast fall the up air with the later hitboxes of the move. A hitbox below the kick arc should be an assist to help land the move more consistently from up close, not something you should be able to abuse.

Riah321:
That is for up air btw.

black-calculus:
Changelog:
[Bayonetta Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8748501/Bayonetta.Changelog.txt)

black-calculus:
Changelog;
[Bayonetta Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8748508/Bayonetta.Changelog.txt)

PR #578, opened 2022-05-12
Daisy changes

black-calculus:
Big batch of Daisy changes for the time being. Will post the changelog and link to the updated romfs once those are up.


black-calculus:
Changelog:
[Daisy Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8726969/Daisy.Changelog.txt)

SkewedAskew:
- Changes feel great. One thing is that the last hit of dair is +22 at 0%, allowing the move to easily chain into more dairs. SDI can mitigate this somewhat already, but the hitstun is still a bit much. I would suggest lowering this to around +16/+17. The first three hits feel fine
- With the increased startup and decreased overall activity of Daisy Bomber, the move feels much clunkier. Would it be acceptable to give it the ability to edge cancel at the end of the landing animation?

PR #582, opened 2022-05-12
Falco changes

black-calculus:
Big batch of Falco changes for the time being. Will post the changelog and link to the updated romfs once those are up.


black-calculus:
Changelog:
[Falco Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8720853/Falco.Changelog.txt)

black-calculus:
Changelog:
[Falco Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8721013/Falco.Changelog.txt)

yoonikoo:
Falco changes overall feels really good. Laser nerfs were needed and yet Falco still feels really stable. Nair is a better option out of shield and fair has become a better drag down and combo tool. Bair nerfs are insignificant and it feels fair to have a weaker weak hit. The jump crouch animation increase has to be the best change overall (I wouldn't even call it a nerf.) As a Melee/P+ player, it feels more natural to do things later than Fox as the jump crouch frame differences really set them apart. The shine readjustments really help with Shine-OOS options and helps to apply pressure using multi-shines. As it was being discussed and people were complaining about the multi-shines, I feel as if these changes are good but it is a little bit too easy to get these out. I understand if the ultimate engine prevents making multishining harder, but reducing how + we are on shield with shines can allow for pressure counterplay. I really think that Shine Grabs are not an issue if you are able to reduce advantage on shield. Besides that, the changes are good to go on nightly besides the + on shield.

Riah321:
If possible to change it, multishines should have a 1 frame buffer. If it's not possible then make it less + on shield. Right now you can literally hold down and mash X+B to buffer unlimited multishines. The rest of the changes are very solid and should be pushed to nightly asap after the shine change.

Also I'll paste this here from my fox comment:
The best solution I could come up with to balance easy multishines is to have shieldstun specifically get hit with staling harder than the rest of the move. Shine is then + on shield and your opponent has to guess your mixup (which can of course be multishine). At the same time multishines are increasingly unsafe if abused and force the user to think about when to use it as a shield pressure tool. I think this is a good way to give shines more nuance in a game where inputs are easier.
(@ me on discord with your thoughts in #balance-discussion or something)

SkewedAskew:
Aside from the above shine changes, looks good to me

Ruuufiooooo:
multi-shines seem a bit easy to do (as said before), but everything else seems pretty cohesive and works well. laser nerfs don't seem to effect the character overall. the biggest take away for me is the JS adjustment from 4 to 5 frames, especially w/ Melee being my main game lol. overall good job

PR #585, opened 2022-05-12
Dedede changes

black-calculus:
Big batch of Dedede changes for the time being. Will post the changelog and link to the updated romfs once those are up.


black-calculus:
motion_list changes for Dedede (to be merged in alongside this PR later):
[dedede motion list.zip](https://github.com/HDR-Development/HewDraw-Remix/files/8715260/dedede.motion.list.zip)

black-calculus:
Updated changelog with accidentally omitted changes for jab 1/2 and dash attack:
[King Dedede Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8723276/King.Dedede.Changelog.txt)

Riah321:
- Add a weak body hitbox to jet hammer.

Looks good so far in training mode.
Jet hammer looks great but the trail lasting so long makes it look more active than it is. I propose a body hitbox for 3 frames after the swing on higher charges when he falls over. (or something similar)
Also for reference this move does more damage and kills better than warlock punch at full charge.

PR #586, opened 2022-05-12
King K. Rool changes

black-calculus:
Big batch of K. Rool changes for the time being. Will post the changelog and link to the updated romfs once those are up.


Riah321:
image
Got this on boot.

jobrien97:
I'm personally against the concept of light/heavy tilts, but the PR functions as expected
Hold B to catch crown is also not a great input, it's very easy to accidentally catch crown. Perhaps there's another input we can try

techyCoder81:
very hard agree on heavy tilts not making sense 99.9% of the time in the context of smash bros - they are called smash attacks

Riah321:
Any chance all this could be revisited soon? Aside from the heavy tilts I'm pretty curious what else is in here. Also wouldn't hurt anyways since krool is a little overtuned in some areas.

PR #587, opened 2022-05-12
Fox changes

black-calculus:
Big batch of Fox changes for the time being. Will post the changelog and link to the updated romfs once those are up.


black-calculus:
Changelog:
[Fox Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/8720962/Fox.Changelog.txt)

yoonikoo:
Fox changes overall have definitely toned the character down. So much to the point that up-throw really cannot connect into anything unless they don't DI. The 5 frame increase really feels bad and unhealthy for the character as it leads to nothing and with drift DI being a thing, you cannot even get a bair off of it. I suggest that you give back some frames for the move (frames 9-10 instead of 7 or 12.) The other changes that really impact him is the multi-hit angle. After vsing a friend for 2 hours, they were able to SDI out of the VERY FIRST HIT OF FAIR (CONSISTENLY.) Besides these two changes, every other change in the log is healthy and makes the character harder to play; this was achieved by decreasing his jump squat frame data. With this, players have to input moves in faster in order to get wave-shines and anything out of it. Again with the shine, refer back to the comment made about Falco in the Falco PR.

Riah321:
- Reduce up throw hitstun and recovery frames
- Fix Forward air multihits.
- Increase shine shieldstun stale scaling.

Changes are solid. Up throw change is so that you can still convert but you have to read DI harder as currently holding out gets you out for free.
Fair has the same issue of that holding out gets you out for free.

The best solution I could come up with to balance easy multishines is to have shieldstun specifically get hit with staling harder than the rest of the move. Shine is then + on shield and your opponent has to guess your mixup (which can of course be multishine). At the same time multishines are increasingly unsafe if abused and force the user to think about when to use it as a shield pressure tool. I think this is a good way to give shines more nuance in a game where inputs are easier.
(@ me on discord with your thoughts in #balance-discussion or something)

PR #590, opened 2022-05-13
CSS, control, and random changes

blu-dev:
Random:
- You see the fighter before the match starts
- Holding Z enables traditional random

CSS:
- Character is preserved across each match

Control:
- New shorthop/footstool button option in control settings


Angstdiscord:
A small bug exists related to cpu levels. It doesn't seem to actually take into account the CPU's actual level. even if it says CPU 9, it will often default to a much lower CPU level in game (either 3 or 2 it seems?) I think this is related to the code that remembers the last fighters.

Angstdiscord:
In addition, it seems to only use the first name of the character when selecting. 

I.e. if it picks wendy, it will still say "Bowser Jr."

Angstdiscord:
Also seems that yink C00 loads C01?

PK-Tone:
There's a lot of settings at the CSS that the game is ignoring: it ignores your chosen skin, and instead randomly assigns one (changes every match). It ignores the cpu difficulty setting (seems to default to level 2-3). If someone is selects "Random" at CSS, the game will assign them a random character, and when the next match starts, they will start with that same previous character again (rather than defaulting to "Random" again).

As for the short-hop/footstool button: I'm so glad to see some new control options! I understand why you wouldn't want the button to be able to double-jump, but if it were possible to buffer an instant double-jump during jumpsquat with this button, it would be absolutely PERFECT for my use-case.

techyCoder81 (approved):

PR #592, opened 2022-05-16
Revert long jab lock animation KB threshold to vanilla

### LINKED ISSUE #530: Re-re-readjust jab locks
Restore vanilla's knockback thresholds for the short and long flop animations, and disable the ability to getup attack out of downdamage

black-calculus:
Resolves #530 .

Pushes the long jab lock threshold down from 999 units of KB to 15 like in vanilla (commit name on original branch is inaccurate).

### DIFF
--- a/romfs/source/fighter/common/param/battle_object.prcxml
+++ b/romfs/source/fighter/common/param/battle_object.prcxml
@@ -6,7 +6,6 @@
   <float hash="damage_level3">40.5</float>
   <float hash="damage_speed_mul">0.03</float>
   <float hash="damage_angle_ground_max">44</float>
-  <float hash="down_damage_s_reaction">999</float>
   <float hash="damage_fly_quake_l">90</float>
   <int hash="damage_fly_speed_up_reaction_frame_min">998</int>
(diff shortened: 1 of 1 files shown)


jobrien97:
Hesitant to approve this until the longer flop animations can get sped up globally. Otherwise we basically have old jab locks + roll option

Riah321:
Works and stuff. Better than having current jab locks since they're almost worse than just letting them miss the tech normally. I say ship it until it can be touched up.

black-calculus:
I'm with Riah here, I'm going to merge these in as I think they'll be much better than what we currently have punish-game wise. Right now jab resets aren't worth going for over just a regular missed tech punish in most scenarios which just makes them pointless for the most part. Old jab locks were only really problematic because they forced a neutral getup which gave you an extra 60+ frames on top of the long flop animation for a guaranteed punish. Now you only have the duration of the flop animation to get a guaranteed punish, and if you want to set up a stronger punish you'll have to read their getup option.

Riah321:
Yo more game depth. Let's go!

PR #599, opened 2022-05-20
Small Quick Draw instakill charge nerfs

black-calculus:
Increase max Quick Draw charge hold time 165F -> 180F, increase required charge amount for instakill 100F -> 160F


Riah321:
While you're at it, when he's hit out a move that uses the blue flames they stick and they're not always stuck to the sword either. Side special works a second time in the air. This is semantic, but shouldn't his crazy air approaching come from his side special JC? Right now he can just.... Jump at you.
Issue has been submitted :)

black-calculus:
> While you're at it, when he's hit out a move that uses the blue flames they stick amd they're not always stuck to the sowrd either. Side special works a second time in the air. This is semantic, but shouldn't hos crazy air approaching come from his side special JC? Right now he can just.... Jump at you.

Post an issue for the flames bug please.

PR #608, opened 2022-05-23
Reduce dash smash attack window from 6f to 2f

kalomaze:
Reduces the 6f window at the start of a dash where an a input will become a smash attack to 2f


jobrien97 (changes requested):
The dash_s4_frame entry is placed in the wrong spot - it should be in between precede_extension and dash_speed_keep_frame

kalomaze:
Fixed

jobrien97 (approved):

PR #618, opened 2022-05-28
Little Mac - raise ledgegrab box

### LINKED ISSUE #412: Mac's up special grab box needs an adjustment.
Mac's grab box needs to be shifted upwards. so it's actually overs his hand. Cause he'll get to ledge, and not grab. even though his fist is there. And I've already mentioned horizontal drift on recovery moves. Which calc responded to.

jobrien97:
upper bound increased 17.5 -> 21.5

Is now able to grab ledge from as low as this:

2022052721272200-0E7DF678130F4F0FA2C88AE72B47AFDF

Resolves #412


Riah321:
Does what it says on the tin.

techyCoder81 (approved):

PR #620, opened 2022-05-31
fixing docker image and auto builds

techyCoder81:


SkewedAskew:
Played for several hours, experienced no issues. I did get a crash when starting a match after lingering on the CSS for about ten minutes or so, but that may be a separate matter. Looks good to me

H3ra-57:
Tested for a couple of hours offline alone and with a friend, have had no crashes so far. Lgtm

blu-dev (approved):
lgtm

PR #625, opened 2022-06-02
Duck Hunt - remove ability to fire Gunman when dead

jobrien97:
Removes ability to fire a Gunman from time between KO -> respawn


Riah321:
Technically works, except that the gunman will still fire after the timer is up. Maybe make them act as if they took damage when DHD dies?

techyCoder81 (approved):

PR #626, opened 2022-06-03
Pivot adjustments

### LINKED ISSUE #531: Pivot readjustment
Restore the input window for pivots to 3F and remove the speed addition during the transition to turn

jobrien97:
- Reinstated late (f3) pivots
- Perfect (sliding) pivots are now only possible when pivot is input within the first 2 frames of dash (identical window as Sm4sh)
- Pivots input after frame 2 of dash are now Melee-style (no slide)

https://user-images.githubusercontent.com/47401664/171768980-d52e21ca-a45e-4619-8b15-dd275d3f1a0d.mp4

Resolves #531

### DIFF
--- a/fighters/common/src/general_statuses/dash.rs
+++ b/fighters/common/src/general_statuses/dash.rs
@@ -506,6 +506,12 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
     && (!ControlModule::check_button_on(fighter.module_accessor, *CONTROL_PAD_BUTTON_CSTICK_ON) || ControlModule::check_button_trigger(fighter.module_accessor, *CONTROL_PAD_BUTTON_CSTICK_ON))  // AND cstick is off, other than the first frame of its input
     && is_backdash_input {  // AND is a backdash input
         //println!("transition to backdash");
+        if fighter.global_table[CURRENT_FRAME].get_i32() <= 2 {
+            VarModule::on_flag(fighter.battle_object, vars::common::CAN_PERFECT_PIVOT);
+        }
+        else {
+            VarModule::off_flag(fighter.battle_object, vars::common::CAN_PERFECT_PIVOT);
+        }
         VarModule::on_flag(fighter.battle_object, vars::common::IS_SMASH_TURN);
         interrupt!(fighter, FIGHTER_STATUS_KIND_TURN, true);
     }
@@ -562,9 +568,6 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
         VarModule::on_flag(fighter.battle_object, vars::common::IS_LATE_PIVOT);
         PostureModule::reverse_lr(fighter.module_accessor);
         PostureModule::update_rot_y_lr(fighter.module_accessor);
-        fighter.clear_lua_stack();
-        lua_args!(fighter, FIGHTER_KINETIC_ENERGY_ID_CONTROL);
-        app::sv_kinetic_energy::unable(fighter.lua_state_agent);
         interrupt!(fighter, FIGHTER_STATUS_KIND_TURN, true);
     }
 
@@ -671,6 +674,10 @@ unsafe fn status_end_dash(fighter: &mut L2CFighterCommon) -> L2CValue {
 	if StatusModule::status_kind_next(fighter.module_accessor) != *FIGHTER_STATUS_KIND_RUN {
 		VarModule::off_flag(fighter.battle_object, vars::common::ENABLE_BOOST_RUN);
 	}
+    if StatusModule::status_kind_next(fighter.module_accessor) != *FIGHTER_STATUS_KIND_TURN {
+        println!("can pp off");
+        VarModule::off_flag(fighter.battle_object, vars::common::CAN_PERFECT_PIVOT);
+    }
 	
 	VarModule::set_float(fighter.battle_object, vars::common::CURR_DASH_SPEED, initial_speed);
 
--- a/fighters/common/src/general_statuses/dash.rs
+++ b/fighters/common/src/general_statuses/dash.rs
@@ -552,15 +552,20 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
 
     interrupt_if!(fighter.sub_ground_check_stop_wall().get_bool());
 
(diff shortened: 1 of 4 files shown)


Riah321:
Sounds like you're removing melee pivots instead of adding them.

jobrien97:
> Sounds like you're removing melee pivots instead of adding them.

Fixed, thanks

Riah321:
Do not merge this change until people argue more. It has a massive effect on a lot of character's ability to play the game in a negative way. A lot of people who use smash 4 pivots do not like it. A common argument is that the change reduces freedom of expression.

  Most people, regardless of if it could be implemented, are agreeing on:

- Pivoting during the first 3 frames of initial dash makes you slide ala smash 4
- Pivoting any time after this during the initial dash halts your momentum ala melee

techyCoder81:
awesome change ngl

fighters/common/src/general_statuses/dash.rs
```diff
@@ -552,15 +552,20 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
 
     interrupt_if!(fighter.sub_ground_check_stop_wall().get_bool());
 
-    if fighter.global_table[CURRENT_FRAME].get_i32() == 0 && stick_x.abs() >= dash_stick_x {
-        // apply speed on f1 of dash (takes effect on f2 ingame)
-        let prev_speed = VarModule::get_float(fighter.battle_object, vars::common::CURR_DASH_SPEED);
-        let applied_speed = (dash_speed * PostureModule::lr(fighter.module_accessor)) + (stick_x.signum() * ((run_accel_mul + (run_accel_add * stick_x.abs())))) + prev_speed;  // initial dash speed + 1f of run acceleration + previous status' last speed
-        //println!("Changing current dash speed: {}", applied_speed);
-        let applied_speed_clamped = applied_speed.clamp(-run_speed_max, run_speed_max);
+    // f3 perfect pivots
```
techyCoder81, line 561: might be good to change this to not say perfect pivot if perfect pivot is removed

IRBDaUrges:
Personally, I dislike this change. The usage of sliding pivots honestly allows for more creative gameplay as well as some characters inherently get worse due to the none slide from the pivots. Bouncing off of what Riah stated. I believe that if you can work initial pivots to have a slide or make it where if you do it frame 3 you get melee pivots is a lot more effective.

harmakanna:
Ultra disagree with removing perfect pivots. The selling point of this game to me was taking from and respecting every smash game and perfect pivot is like, the exemplar for that, so I didn't even think this would ever be on the chopping block. The sliding makes both offense and defense more interesting by being able to move into place during the startup of a move, giving moves like Inkling dtilt more utility and, IMO, making them more fun. Removing it really would just suck to me ¯\\\_(ツ)_/¯

yoonikoo:
This change ultimately will hinder the freedom of expression. The sliding pivots allowed characters to essentially extend combos that they normally are unable to get, allowing for hype play for a skill that took time to learn and master. All in all, I think this change punishes players that took the time to learn and master this tech,.

techyCoder81:
> This change ultimately will hinder the freedom of expression. The sliding pivots allowed characters to essentially extend combos that they normally are unable to get, allowing for hype play for a skill that took time to learn and master. All in all, I think this change punishes players that took the time to learn and master this tech,.

This does not "hinder expression" inherently. This may change some specifically advantage state routes, but our advantage state is already 10x stronger than every previous smash game anyway, so thats really a non issue. On this end, its entirely a preference thing. However, this engine is already wildly slippery in every other way, and we very much lack in ways to have precise control over spacing and timing, something this mechanic has provided for years.
Additionally, switching to traditional pivots opens up a certain methodical playstyle which is currently almost unviable in this engine. This style of pivots is more useful for defensive play and precise neutral interactions, which is something we tend to lack at present anyway. The importance of actually having a reliable way to *stop* really can't be overstated, and its something we literally just *don't have* right now.
Also, implying that this takes away skill is pretty inaccurate. This is no easier or harder than the old pivot behavior, but it debatably may be more *relevant*. We're in beta, and mechanics change. This has been a talking point for something like 6 months. At the end of the day, this is a *different* mechanic, and more specifically one that opens up options that don't currently exist (especially spacing precision).

Fiasco3K:
I truly think this is not a good thing to add due to perfect pivots adding a lot more options and creativity to the game in both the neutral and combo aspect.  I think people should take the time to learn how to perfect pivot and see how they feel about it once they get both of them down.

Marleyan:
Personally not a fan of this. More characters benefit from sliding pivots in comparison to traditional ones, in several ways other than just defensive utility. 
Advantage state is already relatively strong here, but sliding pivots allow for more expressive and unpredictable routes during advantage state that helps differentiates one player's punish game from another and keeps the other player guessing. Sliding pivots also allow an abundance of characters different options for approach in neutral; Mac's sliding pivot up tilt and Kirby's sliding pivot up tilt out of dash are the first two that come to my mind. 
Defensively, sliding pivots allow characters to give up a precise amount of space that is not only less committal than dash back/wavedash back (due to shorter length and being actionable sooner), but also give just enough distance between your character and the opponent's to whiff punish upcoming moves that dashing back and WD back just can't accomplish to the same degree.
I do agree the game does feel a bit slippery (although personally pivots don't contribute to this slippery feel imo) like mentioned in an earlier post, so to combat this, traditional pivots should maybe be placed after the F3 window or somewhere during initial dash after this window.  There is a place for melee-esque pivots in the game to smoothen up the gameplay and have a bit more control over your character possibly, but it shouldn't be replacing sliding pivots-which are far more applicable and generally beneficial in one way or another for a plethora of the roster.

kalomaze:
Don't want this merged tbh

fighters/common/src/general_statuses/turn.rs
```diff
@@ -86,6 +92,8 @@ unsafe extern "C" fn status_turn_main(fighter: &mut L2CFighterCommon) -> L2CValu
         if fighter.global_table[STICK_X].get_f32() == 0.0 {
             VarModule::off_flag(fighter.battle_object, vars::common::DISABLE_BACKDASH);
         }
+        // uncomment for pivot boosts
```
techyCoder81: for reference, is this the block which would add artificial speed within a small window (first few frames only) which could properly be called "perfect pivot"?

techyCoder81 (dismissed):

Riah321:
Looks great! Can't wait to try it out.

techyCoder81 (approved):

PR #630, opened 2022-06-03
Rework aerial glide tossing (airdodge cancelled item throws)

### LINKED ISSUE #615: Aerial Glide Toss reworks
Rework AGTs to be P+ style.

- Toss is only available before intangibility
- AGTs refresh airdodge
- AGTs provide diminishing returns on distance

Spawn_32 mentioned he may do this. This is mostly for documentation.

JoaoAlmeida20:
Changes ~~two~~ three important aspects of AGT:
1. You can no longer cancel an airdodge at any time into the item throw. It's now a 3-frame window at the very start of airdodge, frames 1-3.
2. If you cancel airdodge into item throw, you don't burn your use of airdodge, and can use it again without having to land.
3. If you use your airdodge to AGT again, it will have diminishing returns distance and speed wise each time you AGT, eventually carrying none of the airdodge's speed into the item throw

Additionally this PR also increases Toon Link's bomb throw speed multiplier, it was very low to facilitate bomb regrabs, but with this mechanic it allowed him to chain AGTs for free and fly around. Since the throw speed was also too low for when you wanted to actually throw the bomb anyway, it was agreed to increase it after some discussion, instead of reworking AGTs with some sort of staling mechanic.

Discussion: https://discord.com/channels/659964948365049887/950977159088975883/982277295416967168
Toon Link flight before adjusting his throw strength: https://streamable.com/3xhwja
Regular example showing off the new possibilities and limitations: https://streamable.com/cf4jnq

Resolves #615

### DIFF
--- a/fighters/common/src/opff/var_resets.rs
+++ b/fighters/common/src/opff/var_resets.rs
@@ -9,29 +9,28 @@ use smash::app::lua_bind::*;
 use smash::lib::lua_const::*;
 use smash::hash40;
 
-const DAMAGE_STATUSES : &'static [i32] = &[*FIGHTER_STATUS_KIND_DEAD,
-                                            *FIGHTER_STATUS_KIND_REBIRTH,
-                                            *FIGHTER_STATUS_KIND_WIN,
-                                            *FIGHTER_STATUS_KIND_LOSE,
-                                            *FIGHTER_STATUS_KIND_ENTRY];
+unsafe fn var_resets(boma: &mut BattleObjectModuleAccessor) {
+    let damage_statuses = &[*FIGHTER_STATUS_KIND_DEAD,
+                                        *FIGHTER_STATUS_KIND_REBIRTH,
+                                        *FIGHTER_STATUS_KIND_WIN,
+                                        *FIGHTER_STATUS_KIND_LOSE,
+                                        *FIGHTER_STATUS_KIND_ENTRY];
 
-const DEATH_STATUSES : &'static [i32] = &[*FIGHTER_STATUS_KIND_DAMAGE,
-                                            *FIGHTER_STATUS_KIND_DAMAGE_AIR,
-                                            *FIGHTER_STATUS_KIND_DAMAGE_FLY,
-                                            *FIGHTER_STATUS_KIND_DAMAGE_FLY_ROLL,
-                                            *FIGHTER_STATUS_KIND_DAMAGE_FLY_METEOR,
-                                            *FIGHTER_STATUS_KIND_DAMAGE_FLY_REFLECT_LR,
-                                            *FIGHTER_STATUS_KIND_DAMAGE_FLY_REFLECT_U,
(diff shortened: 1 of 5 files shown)


Riah321:
It works as intended! Really happy with how Young Link's recovery is effected. If your character has a zair and you miss the AGT you'll get that instead which is pretty punishing if you don't want it. Love how it raises the skill floor and ceiling of AGTs.

Do keep in mind that it does open up for some high execution stalling techniques as link and toon link (though not infinite in toon link's case) in some matchups and on certain stages. Adding diminishing distance after a certain number of AGTs, or only allowing for a set amount per airtime would prevent these exploits. If implemented I recommend 3 AGTs before diminishing, or 5 before running out.

After preventing the stalling techniques it's good to go!

techyCoder81 (approved):
short and simple

techyCoder81 (changes requested):
actually i just remembered that we wanted this to be changed so that it can only be done once per airtime, my b

Riah321:
The limit to one per airtime has accidentally made it 0 per airtime. This is rather unfortunate.


Jokes aside you can only get Z-Air out of airdodge and in frame by frame you get nothing. Though I'm sure it'll be fixed with the diminishing returns implementation.

Riah321:
It works! I was going to complain about the AGT scaling but it's exactly where it needs to be! Ship it as soon as you're able.
The link stalls are fixed without making it unusable for normal use. Hooray!

techyCoder81 (dismissed):
Lgtm, nice cleanup. Just need agt retested as well as a small quick check to make sure the other var resets still work (should be fine, logic is sound)

fighters/common/src/opff/var_resets.rs
```diff
@@ -9,156 +9,97 @@ use smash::app::lua_bind::*;
 use smash::lib::lua_const::*;
 use smash::hash40;
 
+unsafe fn var_resets(boma: &mut BattleObjectModuleAccessor) {
+    let damage_statuses = &[*FIGHTER_STATUS_KIND_DEAD,
```
jobrien97: This array should be death_statuses

fighters/common/src/opff/var_resets.rs
```diff
@@ -9,156 +9,97 @@ use smash::app::lua_bind::*;
...
+                                        *FIGHTER_STATUS_KIND_REBIRTH,
+                                        *FIGHTER_STATUS_KIND_WIN,
+                                        *FIGHTER_STATUS_KIND_LOSE,
+                                        *FIGHTER_STATUS_KIND_ENTRY];
+
+    let death_statuses = &[*FIGHTER_STATUS_KIND_DAMAGE,
```
jobrien97: This should be damage_statuses

Riah321:
It works! The tapering of AGTs is in a great spot, you don't get it back if you mash throw after an airdodge completes, and tink cannot fly. Ship it!

Riah321:
Crashes on the character select screen for multiplayer. This could be a nightly issue, however.

techyCoder81 (approved):

PR #633, opened 2022-06-03
fixing buffered aerials oos and adding standing turnaround grab

techyCoder81:
1. This makes it so that if you attempt to buffer an aerial oos, but do not let go of shield (like you can in melee, minus the buffer), you will correctly get an aerial, rather than a jump cancel grab (which is obnoxious).
2. jump cancel grab oos is no longer possible.
3. if you press backwards during the first 5 frames of standing grab, you will get a standing grab facing in the opposite direction. This is similar to how pivot grabs work - if you input a dash grab and then press backwards within the first 5 frames, you transition into pivot grab instead. This works similarly. If you input a standing grab, and then input backwards within the first 5 frames of startup, you get a backwards facing standing grab.

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -139,6 +139,7 @@ pub mod vars {
         pub const JUMP_NEXT: i32 = 82;
         pub const IS_JAB_LOCK_ROLL: i32 = 83;
         pub const SHOULD_TRUMP_TETHER: i32 = 84;
+        pub const DID_TURNAROUND_GRAB: i32 = 85;
         
 
         // int
--- a/fighters/common/src/general_statuses/jumpsquat.rs
+++ b/fighters/common/src/general_statuses/jumpsquat.rs
@@ -100,14 +100,22 @@ unsafe fn status_JumpSquat_Main(fighter: &mut L2CFighterCommon) -> L2CValue {
     }
 
     // begin testing for transitions out of jump squat
-    let situation_kind = fighter.global_table[SITUATION_KIND].get_i32();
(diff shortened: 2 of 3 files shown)


MadVillainLordQ:
The PR mostly works, however, I have a major concern with its implementation as it allows for the grabber to turn around grab in shield. This does a lot to change the dynamic of cross up aerials as one of the main reasons why an opponent would go for those is to avoid being grabbed in the first place. This style of change would remove that aspect and inherently make grabbing/shielding much stronger than I think you intended. Also for consistency sake, I think that the window for the turn around should be 4 frames instead of 5 as that is also the window to turn around during a dash grab. 


UPDATE:

- If you hold grab on any frame that isn't the last frame of your jump squat, it'll result in a buffered airdodge. This is quite strange as grab (Z button) should result in an attack. (Works appropriately for Fox)

- There seems to be a bug when when pressing grab on the last frame of your jump squat while holding shield and jump. For some reason, you will take an extra frame to jump than what is normal. This error seems to be random on the characters it occurs. I've gotten this issue so far with chrom, gnw, roy, Zss, and sheik. Also potentially related is that in the air, the move that is buffered is an aerial, not an airdodge like mentioned above. (Note these characters have 4 frame jump squats, Fox did not have this issue when I tried him)

- On the other hand, Luigi, D.Pit, and Samus haven't had this issue but for some reason under the same conditions, if you press grab on the third frame of their jump squats, they will grab instead. The behavior is inconsistent though and I can't figure out why. They also don't exhibit the behavior above in which if you hold grab on the last frame of jump squat they will perform an aerial. Instead, they will do an airdodge. 

- Lucas can actually grab on the 4th frame of his jump squat if you input grab on that frame. Just like above, the behavior is inconsistent.

Discord Links showing bugs:
https://discord.com/channels/659964948365049887/950979417004466196/982940711638945863

techyCoder81:
Closing this PR as it is being bundled with wuboys jumpsquat refactor instead

PR #636, opened 2022-06-04
Removed Attack Cancel JC Grab

### LINKED ISSUE #635: Remove the ability to cancel into Grab during an Attack Cancel.
Imagine someone running up to you
Up Smash
Attack Cancel
Cancel the Attack Cancel into Grab
that's just silly.

WuBoytH:
Resolves #635


MadVillainLordQ:
Tested in game and is working as intended. The interesting tidbit is that airdodge is buffered when doing AC Grab. This is strange behavior as Grab in the air is supposed result in an attack, not airdodge. I believe this is due to how Z - attack was coded rather than any incorrect implementation on your end.

techyCoder81 (approved):
Looks mostly simple enough to me, seems to touch a bunch of js logic

PR #638, opened 2022-06-04
Revert dash_s4_frame to 6f

kalomaze:
It has strange bugs, most prominently with auto-turnaround characters (shotos, terry, kazuya)


techyCoder81 (approved):
small change, seems reasonable

PR #642, opened 2022-06-05
Charizard Dtaunt windbox size reduction

MadVillainLordQ:
Charizard's current Dtaunt has a massive windbox that makes it an effective edge guarding tool. This coupled with the fact that it lasts for 20 frames, makes this a very non interactive edge guarding tool that is a bit ridiculous on this character. I will test in game to make sure the new windbox size and placement makes sense as I'm more or less taking an educated guess at its placement.

Old
2022060500374500-0E7DF678130F4F0FA2C88AE72B47AFDF

New
2022060504172700-0E7DF678130F4F0FA2C88AE72B47AFDF


techyCoder81 (approved):
Extremely simple acmd change, imo no hard testing required. Existing documentation is sufficient.

PR #643, opened 2022-06-05
Allow runbrake to teeter cancel

JoaoAlmeida20:
Just another status that makes sense to be able to go into teeter, and as it did in Melee.
Brief discussion about this: https://discord.com/channels/659964948365049887/950977159088975883/982409581210787901


JoaoAlmeida20:
Tested it myself and it's working as intended now: https://streamable.com/4521jv
To explain the condition removed, it was a leftover from when there was no list of statuses checked and anything with GROUND_CORRECT_KIND_GROUND could be teeter cancelled which led to lots of non sensical situations particularly with special moves. On the other hand, run brake isn't a status with GROUND_CORRECT_KIND_GROUND so it was now preventing it from working properly. And it didn't make sense to give it that ground correct kind as we don't want run brake to slide off edges

Riah321:
Does what it said on the tin.

jobrien97 (approved):
lgtm

PR #649, opened 2022-06-07
Isabelle Recovery - Version 2

HisashiASM:
Power is stored to a threshold and some other stuff, see below.

This is the balloon trip cancel (btc), it saves your recovery power and deducts 100 power from the saved value.
https://user-images.githubusercontent.com/98564957/172276307-39c32289-768d-460f-8c56-06f765d03d38.mp4

This is the balloon trip quick cancel (btqc), it drains all of your recovery for a bonus airtime... find out how to use it!
https://user-images.githubusercontent.com/98564957/172276029-bccdb5c8-8241-44c2-a6c8-d41eaffcbc8c.mp4

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -391,6 +391,14 @@ pub mod vars {
         pub const AIR_CHAIN_COMBO_NUM: i32 = 0x1003;
     }
 
+    pub mod shizue {
+        // flags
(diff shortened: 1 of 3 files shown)


Riah321:
Looks good to me. I like the effects :)
Can't wait for this to be shipped so we can stop worrying about it lol.

techyCoder81 (approved):

PR #661, opened 2022-06-09
Update Air Tether input logic

### LINKED ISSUE #660: Update Air Tether Input Logic
Currently the game uses a single CMD_CAT check against `FIGHTER_PAD_CMD_CAT2_FLAG_AIR_LASSO`. This cat flag will become active if you are holding the Shield button and press Attack, which means that even if you are "no longer buffering Shield" (it's been at least 5 frames since you've started holding it, or your control buffer has been cleared manually), you will still get Air Tether.

This aims to change the logic for inputting Air Tethers so that the Shield input needs to still be within the buffer to input Air Tether.

WuBoytH:
Resolves #660


jobrien97 (dismissed):
lgtm
Surely there'll be no issues this time right? :^)

techyCoder81 (approved):

Riah321:
LGTM

PR #662, opened 2022-06-09
Small Dr. Mario changes

kalomaze:
CHANGES:
- Nair effects tweaked to better match the framedata
- Nair sweetspot on f10 instead of f11
- Nair given electrical SFX on the sweetspot
- Chill Pill KBG increased 60 -> 90
image

PR also on the private repository to change his nair animation to pop out at the sweetspot: https://github.com/HDR-Development/hdr-private/pull/108


Riah321:
Looks good to me. Chill pill still doesn't even kill training mode ridley on the top platform of battlefield at 130%. Pill is a little wonky but it's good enough. Fix chill pill and then ship it :)

kalomaze:
Pill param changes reverted per request of Madvillain, should be good to go

techyCoder81:
this is just waiting on a final functional test

Riah321:
It's good to go. Ship it :)

fighters/mariod/src/acmd/aerials.rs
```diff
@@ -50,6 +50,21 @@ unsafe fn mariod_attack_air_n_effect(fighter: &mut L2CAgentBase) {
     }
 }
 
+#[acmd_script( agent = "mariod", script = "sound_attackairn" , category = ACMD_SOUND , low_priority)]
+unsafe fn sound_attackairn(fighter: &mut L2CAgentBase) {
+    let lua_state = fighter.lua_state_agent;
+    let boma = fighter.boma();
```
jobrien97, line 56: Are you not keeping the vanilla "se_common_punch_kick_swing_s" sfx call on frame 3?

kalomaze: I figured it'd make more sense to call just some electric sfx for the sweetspot so I didn't bother dumping/getting the vanilla sound script first, but if you think both should be there then that's something I can do

jobrien97: > I figured it'd make more sense to call just some electric sfx for the sweetspot so I didn't bother dumping/getting the vanilla sound script first, but if you think both should be there then that's something I can do

Ye it comes from the initial kick (frame 3). Just sells it a bit more

kalomaze: Added

techyCoder81 (approved):

PR #671, opened 2022-06-12
Shine corrections/ changes on block frame advantage

### LINKED ISSUE #670: Shine on block Frame Advantage Readjustment
With the removal of character's being unable to grab during blockstun, a few moves have to readjust in terms of their frame advantage:

New Proposed Down Special Frame Advantage:

Fox - +1
Wolf/Dpit - +2
Falco/Zelda - +3
Lucas/Ness - +4

Old Down Special Frame Advantage:

Fox/Wolf/Lucas/Ness - +1
Dpit - +2
Zelda - +3
Falco - +4

First off, the main determining factor of the shine frame advantage is that shine to jc grab will always lose to roll, so no shine like move will ever be above +4. 

- Next, is a combination of accessibility and jump squat. Fox/Wolf/Falco shines are ordered in this fashion due to the differences in jump squat and reward off of grab. Generally fox usually has the best reward, then wolf, and then falco. 

- Dpit is left in the +2 (could honestly be +1/0, we're most likely gonna remove his shine at some point as it isn't necessary in his tool kit) as his shine comes out slower than the rest at frame 5.

- Zelda follows a similar logic to Dpit and she is placed in the +3 category as her shine comes out frame 8. She does have some benefits in comparison to the others as her shine does have a significantly larger hitbox along with intangibility on frames 5-8.

- Lucas magnet is frame 5 like dpit so one would think it makes sense for him to be at +2 like Dpit. However, unlike the other characters on the list, he has a tether grab that is active on frame 14 so the threat of grabbing from his magnet is not nearly as frame tight as the other characters thus affording him leeway on his frame advantage. We also have to take into consideration that he is a DJC character, so a lot of his pressure comes from doing magnet -> djc aerial stuff which is very strong but he wouldn't be able to straight up grab you from it. So he has more pressure when doing an aerial magnet to DJC aerial. Which is a pretty central point to how he was designed in PM and I'm assuming  that is in the same vein of the design direction HDR is taking with the character. 

- With Ness, I'm very reluctant to place him at +4. While you can avoid the grab with a frame perfect roll (in reality with tap buffer, the window is 5 frames), it would be reasonable to give the defender a two frame leeway in avoiding the grab in these scenarios. The reason Ness was given the +4 was due to his magnet being slower than the rest at 7 frames. Also among the characters in the list, he would most likely have the hardest time actually getting magnet on shield against his opponents due to the combination of his speed and aerial frame data. If his magnet were to be improved to frame 5, then the case can be made to make it +3. But as it stands, it doesn't make much sense to put it in the same frame advantage range as the other characters whose shines are generally much more applicable for shield pressure.

MadVillainLordQ:
Resolves #670

In addition to resolving the above issue, Falco and Wolf's shines have had their active hitboxes on their shines reduced to 1 frame. Previously, it was 2 frames for falco (which extended the hitboxes size due to interpolation that occured), and Wolf was 3. Ness magnet also received an increase in damage (with kb compensation), to avoid having to increase the shield stun multiplier by too much. (This can be reverted if the damage buff proves to be too strong in other aspects of his gameplay)

What's missing:

Lucas shine is still +3 instead of +4. This will be resolved once his jumpcancel frame gets improved in a future patch.


PK-Tone:
Regarding Ness's magnet: with fixed-knockback moves, there is no such thing as kb-compensation for increased damage; the move's damage does not factor into the knockback equation in any way. If you look at the knockback equation with regards to fixed knockback attacks, "p" (the opponent's percent) is always substituted by 10 (i.e. the move always assumes the opponent is at 10%), and "d" (the move's raw damage) is substituted by the move's FKB.

If you want to buff the shieldstun but leave on-hit advantage intact, just increase the damage and leave the knockback data as-is (or you could just try adding a shieldstun modifier). Reducing the knockback as "compensation" for increased damage is just a straight-up nerf to hitstun and combos.

PR #675, opened 2022-06-14
Lucas - reinstate frame-perfect DJCZ

jobrien97:
zair input logic reworked once again!!


WuBoytH (changes requested):
The only thing I would do is store all of those checked values into variables instead so it looks less cluttered

WuBoytH (approved):
lgtm

PR #682, opened 2022-06-15
metaquick fx and remove lots of println

blu-dev:
https://cdn.discordapp.com/attachments/950979381654855701/986704676361764914/2022-06-15_11-51-22.mp4

### DIFF
--- a/fighters/metaknight/src/meta_quick.rs
+++ b/fighters/metaknight/src/meta_quick.rs
@@ -151,9 +151,6 @@ pub unsafe fn run(fighter: &mut smash::lua2cpp::L2CFighterCommon) {
     MeterModule::watch_damage(fighter.object(), true);
     MeterModule::set_damage_gain_mul(fighter.object(), 6.0);
     
-    //println!("Meter Module: {}", MeterModule::meter(fighter.object()));
-    //println!("Gimmick Timer: {}", VarModule::get_int(fighter.object(), vars::common::GIMMICK_TIMER));
-
     // logic for starting metaquick:
     // if our meter is >= 10, then we want to check for taunt
     // if we are pressing down taunt, then we want to enter metaquick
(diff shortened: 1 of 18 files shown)


fighters/metaknight/src/meta_quick.rs
```diff
@@ -2,34 +2,219 @@
...
+    // logic for starting metaquick:
+    // if our meter is >= 10, then we want to check for taunt
+    // if we are pressing down taunt, then we want to enter metaquick
+    // if we are in the air and in an actionable state, we want to enter the metaquick summon status
+    // if we are grounded and in a state where we would normally enter down taunt, we want to enter
+    // the metaquick summon status.
+    // this should be achievable by not draining the meter if we are about to enter down taunt
+
+    if MeterModule::level(fighter.object()) >= 10 {
+        if fighter.is_cat_flag(Cat2::AppealLw) {
```
techyCoder81, line 163: any particular reason why this is AppealLw instead of AppealAll like before?

fighters/metaknight/src/meta_quick.rs
```diff
@@ -2,34 +2,219 @@
...
+            handle_start_metaquick(fighter);
+        } else if VarModule::get_int(fighter.battle_object, vars::metaknight::META_QUICK_CHARGE_EFFECT_HANDLE) == -1 {
+            VarModule::set_int(
+                fighter.battle_object,
+                vars::metaknight::META_QUICK_CHARGE_EFFECT_HANDLE,
+                EffectModule::req_common(fighter.module_accessor, Hash40::new("charge_max"), 0.0) as i32
+            );
+        }
+        // return;
```
techyCoder81: honestly at this point all of this commented stuff should just be removed, theres no reason to keep it. If its an issue we would just revert, and if not, its just clutter. So imo just delete it (at least once tested).

fighters/metaknight/src/meta_quick.rs
```diff
@@ -2,34 +2,219 @@
 use super::*;
 use globals::*;
 
+/// pre status for metaquick summon
+/// handles initialization
+pub unsafe extern "C" fn metaquick_summon_pre(fighter: &mut L2CFighterCommon) -> L2CValue {
```
techyCoder81, line 7: my only question here is, do we want to restructure this to have this structure instead, to be a little more organized:
```
/meta_quick/
    /status.rs
    /acmd.rs
    /opff.rs

techyCoder81 (changes requested):
overall good besides some preferred changes

fighters/metaknight/src/meta_quick.rs
blu-dev: 1. the down taunt animation is close enough to the animation that we play for metaquick
2. by mapping all taunt buttons, you are removing the ability to footstool in the air without using the footstool button

fighters/metaknight/src/meta_quick.rs
blu-dev: yep I just forgot to remove them, will do so

fighters/metaknight/src/meta_quick.rs
blu-dev: I don't think we need to change the structure. The status code is small enough that putting it in it's own file would be wasteful imo. The ACMD is also really small so I don't think it's important enough to do that. The bulk of the speed setting and management logic is in OPFF anyways.

techyCoder81:
ready for final testing

Riah321:
Tested and works! Fair trains aren't free either, if that was a concern.
Ship it :)

techyCoder81 (approved):

PR #683, opened 2022-06-15
fixing Shorthop-Button-cancelling behavior

techyCoder81:
Making sure jumps using shorthop are locked behind the appropriate transition term.


ethangarsia517:
This appears to have fixed aerial attacks having 0 frames of landing lag when using Shorthop button right after. Aerial attacks are no longer auto canceled when using the Shorthop button right after.

PR #692, opened 2022-06-16
Joker - Temporarily remove gun beam gfx

jobrien97:
Those were a work-in-progress, to say the least lol

Removes these things:
2022061512592500-0E7DF678130F4F0FA2C88AE72B47AFDF


blu-dev (changes requested):
could use more comments imo

blu-dev (approved):
ok but fr, yes these need more time in the oven

PR #693, opened 2022-06-16
VarModule Rework

### LINKED ISSUE #659: Rework VarModule to allow for Status Variables.
VarModule currently only has what Smash refers to as "INSTANCE" variables, which persist until they are manually changed. However, Smash also has "STATUS" variables, which are variables that (typically) persist only until they change statuses, in which case they are automatically reset to their default state. They won't _always_ clear, but that's dependent on the `KEEP_FLAG/INT/FLOAT` flag passed into `StatusModule::init_settings` - this will mostly be the case for things like transitioning from GUARD_ON to GUARD to GUARD_OFF, or if a character transitions into unique SPECIAL statuses, in which case the "STATUS" variables will persist.

In reworking VarModule, I will also be going through and changing the value of each constant. I will put a guide to all of the changes within `dynamic/src/consts.rs`.

WuBoytH:
Resolves #659


hashbrun555:
I tested with Marth and everything seems to be the same with no crashes.

Riah321:
Completely fucks online. Got a crash during match load, always desynced on death or on hitting each other.

WuBoytH:
> Completely fucks online. Got a crash during match load, always desynced on death or on hitting each other.

Can I get some more information on this? Were you playing against someone with or without the PR? What characters were you picking? Etc

Riah321:
> > Completely fucks online. Got a crash during match load, always desynced on death or on hitting each other.
> 
> Can I get some more information on this? Were you playing against someone with or without the PR? What characters were you picking? Etc

I typed up a reply to this on two separate occasions and then forgot to send it both times. F

Hitting each other, dying, and sometimes anyways would desync immediately. Cross was playing mario and then dark samus and I was playing DK. We both had the PR and the same romfs.

MadVillainLordQ:
Lgtm.
Played a few online matches and everything worked fine.

jobrien97 (dismissed):

Issue #132, opened 2022-02-17
Theoretical options for control remapping ui and logic


techyCoder81:
in theory we can hook when the game opens the controls UI and show a different UI instead, using a web menu. From there we could figure out the logic and work required to:
- [x] make a UI that allows a user to select various options
- [x] store those options in extra space within InputKind fields in control settings/save data
- [x] make new pad flags using space stolen from stick position mapping, which will work over wifi

control options desired:
- [x] dedicated smash button
- [x] dedicated short hop button (use ctrl pad jump mini)
- [ ] footstool button (requires new pad flag - 1 bit)
- [ ] "strong attack"/"smash attacks when pressing Attack?" (requires mapping the setting in pad flag space - 1 bit)
- [ ] stick sensitivity and deadzones for sticks
- [ ] analog stick thresholds
- [ ] mapping the inner button on gcc
- [ ] "should your short hop button only footstool in the air?"
- [ ] fullhop button (requires new pad flag - 1 bit)
- [ ] "do you want a sh macro?" (requires mapping the setting in pad flag space - 1 bit)
total necessary pad flag bits: 4 bits

currently, transmitted data is:
X - x stick position
Y - y stick position
0 - dont care/cant use
A - button space
`XXXXXXXX YYYYYYYY 000000000000000000000000000000 AAAAAAAAAAAAAAAAAA`

instead we could:
X - x stick position
Y - y stick position
0 - dont care/cant use
A - button space
B - new button space to mask into
`XXXXXX YYYYYY BBBB 000000000000000000000000000000 AAAAAAAAAAAAAAAAAA`

we can pack any new pad flags into the existing bits that translate stick position:
16 bits is currently used to store stick, but we can use 12 to create a 64x64 control grid, and use the remaining 4 for the above settings.
- [ ] footstool pad flag
- [ ] "strong attack"/"smash attacks when pressing Attack?" setting
- [ ] fullhop button
- [ ] "do you want a sh macro?"

as far as reading, writing, and storing controller settings:
since `InputKind` only technically uses the first 4 bits of its 8 bit size, we could use the back half of each of these InputKind fields to store additional data, in theory:
https://github.com/HDR-Development/HewDraw-Remix/blob/dev/dynamic/src/ext.rs#L662-L722

Also:
since the game sends pad flag data even during match startup, theoretically we could instead pack the pad flag data with 16 + 18 bits worth of settings data, if necessary during that time.

blu-dev:
[...]

PR #701, opened 2022-06-17
New Controls Options (Footstool, Short Hop, Smash, Tilt)

### LINKED ISSUE #684: Shorthop/Footstool doesn't work when jumping in special scenarios or for Jump Cancelable Special Moves
The Shorthop/Footstool button does not allow special actions that can normally be jumped or short-hopped out of, to shorthop. These scenarios so far include -
   •Donkey Kong's Cargo Throw
   •Bowser Jr.'s Clown Cart, grounded

It also will not work when using Jump Cancelable specials that can be short-hopped out of using the normal Jump button. It would probably be safe to assume all Cancelable Specials are as such, since nothing that I tested worked, but here is what I tested not to work -
   •Samus and Dark Samus' Charge Shot
   •Fox's Reflector
   •Little Mac's Straight Lunge
   •Ike's Quick Draw
   •Squirtle's Water Gun
   •Lucario's Extreme Speed, grounded
   •Mii Swordfighter's Gale Stab

techyCoder81:
This adds:
- Taunt/Footstool button
- Short Hop button
- Smash Attack button
- Tilt Attack button

Tilt attack now has a dash->ftilt leniency window just like fsmash where your first X frames of dash are allowed to transition to ftilt. This uses the same frame window param as the fsmash leniency window.

This should also fix the behavior of the previous Shorthop/Taunt button which made it not function with gimmick jumps such as cargo carry or bjr side special (fixes #684). When testing, please also verify that this is the case.

The romfs changes necessary for this are in the newest nightly.

Related to #132


Riah321:
Crashed while entering and exiting the up or right dpad button menu (I don't remember which) on procons

techyCoder81:
> Crashed while entering and exiting the up or right dpad button menu (I don't remember which) on procons

have not been able to reproduce this behavior.

albertlopes:
Hi @techyCoder81, i'd like to ask your help and permission to try to make a standalone mod of this implemantiton. 

Is it possible to export only those options as standalone mod.

Taunt/Footstool button
Short Hop button
Smash Attack button
Tilt Attack button

if so could you guide me what tools i need to do and if you let I use your work as base. I'm a developer but never moded any game so I even tho I will struglle, eventualy  i will understand the language.

PR #707, opened 2022-06-18
Drift DI scaling adjustments

techyCoder81:
Drift DI now scales down as your percent gets higher, starting from 1.0 and scaling down to 0.75 strength at 100%. It continues to scale lower from there, by 0.25 per 100% damage.
Drift DI when at 0 horizontal knockback is reduced from 0.5 to 0.25


kalomaze:
Played quite a few matches with Skudo, feels a lot more appropriate kill % wise without negatively impacting combo variety. Maybe could be a bit more aggressive, scaling wise, like I think 75% strength at ~80% would make sense

PR #716, opened 2022-06-20
remove c-stick drift

blu-dev:
time to party, bitches

does not yet fix c-stick fastfalling, this will likely be open for a while


WuBoytH (approved):
Tested this in game with blujay and it causes no issues aside from one, which will be looked into later.
LGTM

PR #717, opened 2022-06-20
Hitstun mul reduction 0.45 -> 0.42

techyCoder81:
damage_frame_mul 0.45 -> 0.42

For reference, every previous game except 64 used a value of 0.40


blu-dev (changes requested):
could use more comments imo

PR #720, opened 2022-06-21
Fixed luigi fully invince downb

### LINKED ISSUE #719: Luigi fully invince downb
if luigi enters the air during the startup of downb (during the frames it has invince), then the invince never recieves a cancel command and is fully invince for the entire animation.

This used to be a vanilla bug that i feel like applies to a couple other chars but i cant remember. It appears it was fixed in vanilla but something about the edits in hdr has re-introduced the bug

https://streamable.com/cjuw4q

EonTAS:
adds a command to remove luigi downb invince when in the air so he can't have a permanently invincible downb

fixes #719


WuBoytH (approved):
Nice catch, LGTM

PR #729, opened 2022-06-23
corrin activity adjustments

### LINKED ISSUE #728: Corrin normals be cheating

TeamPearson:
fixes #728 

up air had the last 2 frames of hitbox removed

up tilt has gone from 11 frames of activity to 7


TeamPearson:
fixes #728

techyCoder81:
do these changes correctly account for the effects? If the effects are wack compared to the the hitbox, they should be adjusted as well, just like marth and roy's. Also, it may be wise to look at fair and possibly bair too for these same reasons.

TeamPearson:
up tilt was motion rated and looks normal

up air will need the affect script added and altered to match the hitboxes now

TeamPearson:
> do these changes correctly account for the effects? If the effects are wack compared to the the hitbox, they should be adjusted as well, just like marth and roy's. Also, it may be wise to look at fair and possibly bair too for these same reasons.

with recent update everything should line up visually

TeamPearson:
>nair angles changed from 70 to 50

>down tilt angles adjusted
*outer two hitboxes angles changed from 100 to 65
*inner two hitboxes angles changed from 100 to 75

>upB ledge grab box slightly enlarged to allow for sweetspot

PR #734, opened 2022-06-24
Ledge behavior bugfixes/rework

### LINKED ISSUE #235: Move ledge occupancy logic out of OPFF and back to ledge functions, and change tether reel-ins to transition to trump when they attempt to grab an occupied ledge
In the previous transition from C++ to Rust/Skyline, I moved the ledge grab function logic to the common OPFF to write in a hacky solution to ledge tethers not transitioning to trump when attempting to reel into an occupied ledge, which also affected ledge occupancy timings.

Specifically, due to the main function we used for OPFFs back in the C++ implementation of HDR (get_command_flag_cat) being run 4 times per frame, the game was able to check between frames whether or not to trump you as the tether character in the OPFF which caused no issues. However, now that we are using sys_line_control_system_fighter and the logic only runs once per frame, issues arose where tether characters would not be trumped properly, so I moved the ledge occupancy setting logic to where the OPFF logic for trumping you as the tether character reeling in to the occupied ledge was, and added a distance proximity window so you would be trumped before you actually reached ledge.

Now that we know a bit more about the game, we should place the ledge occupancy logic back into the ledge functions where they belong as well as add functionality to make tether reel-ins transition into ledge trump rather than ledge grab upon attempting to grab a ledge occupied as marked.

This should also resolve the issue where the player with the higher-numbered port will get trumped when multiple characters attempt to grab the ledge at the same time.

Related to issue #232 .

### LINKED ISSUE #808: Remove 2-framing
The first 2 frames of a ledgegrab from below ledge are still vulnerable. This is no longer necessary with our wildly different ledge mechanics.

jobrien97:
Should fix issues with tether characters being able to get trumped when the other character is no longer occupying ledge.
Should also fix same-frame ledgegrabs causing trumps.

Also removes the ability to grab ledge during your runfall/walkfall animation (the first few frames after running off an edge). This is to allow for a more interesting race to ledge rather than simply running off and grabbing, and to prevent accidental ledgegrabs when attempting to edgeguard.

Also removes 2-framing: `FIGHTER_STATUS_KIND_CLIFF_CATCH_MOVE` has been given intangibility.

Resolves #235
Resolves #808


fighters/common/src/function_hooks/change_status.rs
```diff
@@ -6,11 +6,37 @@ unsafe fn change_status_request_hook(boma: &mut BattleObjectModuleAccessor, stat
...
+                        if i == player_number || VarModule::get_float(object, vars::common::instance::LEDGE_POS_X) == 0.0 {
+                            continue;
+                        }
+    
+                        if pos.x == VarModule::get_float(object, vars::common::instance::LEDGE_POS_X) && pos.y == VarModule::get_float(object, vars::common::instance::LEDGE_POS_Y) {
+                            next_status = *FIGHTER_STATUS_KIND_CLIFF_ROBBED;
+                        }
+    
+                        let module_accessor = &mut *(*object).module_accessor;
+                        if module_accessor.kind() == *FIGHTER_KIND_POPO {
```
WuBoytH, line 28: I have a single concern and it's about this line.

I believe the fighter check should be for either POPO or NANA, because there are certain costumes where Nana is the one in the lead. However, I can't verify this on my own, so I ask that you please check.

WuBoytH: I have been told that the lead always reads as POPO

WuBoytH (approved):
LGTM

PR #748, opened 2022-06-27
ECB bugfixes/cleanup

### LINKED ISSUE #756: Character ECBs do not properly offset when buried
due to new ECB refactor

jobrien97:
Fixes crashes when spawning items
Fixes some issues with buries/wavelands

Implementation also cleaned up

Fixes #756


Riah321:
Feels much cleaner and works as intended. Buries might need another pass eventually but for now you should ship it since they will probably be on a per-character basis.

dynamic/src/modules/var.rs
```diff
@@ -113,6 +116,16 @@ pub mod VarModule {
     /// Resets all values
     pub const RESET_ALL:      u8 = 0xFF;
 
+    /// Resets various `VarModule` arrays depending on the mask
```
WuBoytH: I don't know what will happen when you merge this because I believe I've updated the description of this function
check https://github.com/HDR-Development/HewDraw-Remix/commit/4726c7bbee7a01a3b9ef2f74cdca658d175a3262 for the proper description, fix it if you want or just merge it I don't think it's a big deal

WuBoytH (approved):
LGTM

PR #755, opened 2022-06-28
Fix VarModule init_settings crash

### LINKED ISSUE #754: VarModule Rework does not take into account items
The `StatusModule::init_settings` hook was modified to allow for automatic resetting of "status" variables in VarModule. However, because all items do not get created with VarModule (or any of HDR's custom modules), the hook crashes when any item attempts to access it.

To fix this, I've added an exposed function to VarModule that returns if VarModule exists for that BattleObject, and then attempt to reset VarModule only if it's true.

WuBoytH:
Resolves #754

### DIFF
--- a/dynamic/src/modules/var.rs
+++ b/dynamic/src/modules/var.rs
@@ -1,6 +1,9 @@
 use smash::phx::*;
 use smash::app::BattleObject;
 extern "Rust" {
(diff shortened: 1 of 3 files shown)


jobrien97 (approved):
lgtm & tested ingame

PR #760, opened 2022-06-29
Marth shieldstun multiplier and special adjustments

MadVillainLordQ:
Previously during alpha release, a few of marth's tools had shieldstun multipliers set on them to reduce their shield frame advantage. At the time, this sort of change made sense as Marth was widely considered to be among the strongest characters of the game and this was meant to balance the character. Over time with nerfs to his range and hitbox properties, Marth stands to be much more in line with the rest of the cast, and thus we'd like to experiment again with the character not having any shieldstun multipliers. 

Additionally, changes made to his specials:
- Aerial Up b intangibility matches his grounded up b intangibility
- Side B 1 had its length/startup reduced to properly match the animation's visual (Hitboxes in were repositioned slightly to accurately convey its range, startup is now frame 9 instead of 10)

Imgur link: https://imgur.com/a/68OWZjE


jobrien97 (approved):
lgtm

PR #766, opened 2022-06-30
Simon/Shulk Adjustments

Jamillian8:
Reduced damage of fair, bair and f tilt
Reduced width of fair, bair, and f tilt hitbox by half to match vanilla's hitbox
Normalized damage and knockback through out the whip, removed tipper as well for fair, bair, f tilt (similar to Lucina/Chrom)


SkewedAskew:
For Simon, the damage on ftilt/fair/bair seems too low. I would suggest trying these adjustments out:

ftilt: 7.0 -> 9.0
fair: 8.0 -> 10.0
bair: 8.0 -> 11.0

Additionally as previously discussed, experiment with moving the hitboxes inward to reduce the disjoint beyond the tip of the whip

SkewedAskew:
For Shulk, please denote the changes in frame data explicitly in your PR description as raw motion rates can be difficult to decipher. At some point you should change the syntax of your motion rate changes to a more readable form, which one of us can instruct you on

PR #771, opened 2022-06-30
Add RoA wavedash macro + wavedash improvements

### LINKED ISSUE #765: Chaining wavedashes will sometimes cause roll
Inherently due to our buffer system, attempting to chain wavedashes will sometimes be inconsistent and result in a roll.

Will also look into implementing Rivals wavedash macro (jump + shield on same frame).

jobrien97:
New wavedash macro:
- Previously, inputting Jump (tap or button) + Shield on the same frame would result in an empty hop
- Now, Jump + Shield = wavedash

Also disabled shielding/rolling while Jump is input, to allow for easier chaining of wavedashes.

Resolves #765

### DIFF
--- a/dynamic/src/ext.rs
+++ b/dynamic/src/ext.rs
@@ -238,6 +238,7 @@ bitflags! {
 
     pub struct CatHdr: i32 {
         const TiltAttack = 0x1;
+        const Wavedash = 0x2;
     }
 
     pub struct PadFlag: i32 {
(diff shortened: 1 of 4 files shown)


Riah321:
Feels so much better! Can't wait for this to be merged. (also lgtm)

WuBoytH (approved):
Looks good to me, lgtm.
Only thing that might happen later is I may fully implement the rest of the `sub_transition_group_check_` hooks you've added, but again that's not for this PR

PR #778, opened 2022-07-01
Shulk adjustments

Jamillian8:
Increased start up on fair, bair and nair.


Riah321:
Frame data please :) also you can make nair rotate slower if that would suit it better.

PR #780, opened 2022-07-01
Fox - Firefox startup ledgegrab

jobrien97:
Allows you to ledgegrab during Firefox startup, so you can ledgestall like the good ol days


Scrubahh:
Tested on many stages including those with wacky ledges like Lylat and Umbra Clock Tower. I couldn't find any bugs from what I've tried so far. The snap doesn't work if Fox's head is even barely above the ledge, and seems like a pretty narrow window on what grabs in comparison to Melee. But in terms of pure functionality it seems to work as intended, Played on console version.

WuBoytH (approved):
Lgtm

PR #799, opened 2022-07-08
Pikachu - disable Quick Attack Jump Cancel after Quick Attack 2

### LINKED ISSUE #739: Remove Pikachu qa2jc
remove the jump cancel on quick attack 2

jobrien97:
Quick Attack can only be jump canceled when landing after the first warp.

Resolves #739


WuBoytH (approved):
LGTM

PR #800, opened 2022-07-08
Meta knight fair pays taxes 

kalomaze:
50 > 70 bkb
160 > 155 kbg
1.3x damage on all hits
1.2x sdi multiplier on all hits


techyCoder81:
I'm sorry you feel that way

PR #801, opened 2022-07-08
Wavedash macro adjustment

jobrien97:
Previously, Jump + Shield had to be input on the _same frame_ to get a buffered wavedash.

Now, Shield can be pressed at any time within Jump input's buffer window to register a buffered wavedash.


WuBoytH (approved):
As much as I don't like change_status being used to hook and enable flags, there's the age old saying "if it works it works"
LGTM

PR #804, opened 2022-07-08
Add a seperate hitbox cancel to air Luigi Downb

EonTAS:
lines up hitbox timings with grounded downb so that if an air transition occurs the frame it should cancel the event isnt skipped.

extended fix to #719 around hitbox active frames that apparently also has a problem here

example of the issue, also shows the invince bug since i found the issue on beta build 
https://streamable.com/oh41ma


jobrien97:
What really should be done here is standardizing the launcher hitbox's activity so this weird inconsistency doesn't exist.
I would make the aerial version last 3f as well.

EonTAS:
> What really should be done here is standardizing the launcher hitbox's activity so this weird inconsistency doesn't exist. I would make the aerial version last 3f as well.

that makes sense to do, my general goal was just to not touch balance since this appears to be changed from vanilla in some way so I didnt want to make a change beyond just fixing the bug, but will remove the initial cancel to fix this

jobrien97 (approved):
lgtm

PR #818, opened 2022-07-14
Falco - Firebird startup ledgegrab

jobrien97:
Allows you to ledgegrab during Firebird startup, so you can ledgestall like the good ol days


WuBoytH (approved):
Why wasn't this included with Fox's Firefox change?

Either way, logic looks to be the same as Fox's so lgtm

PR #820, opened 2022-07-14
Knockback voice clip adjustments

### LINKED ISSUE #819: Play heavy knockback voice clips immediately on hit
When characters incur heavy knockback, their voice clips do not play until frame 6 of the knockback animation, which is unnaturally delayed and sometimes doesn't even play before they reach the blastzone.
Will change them to play on frame 1, which is identical to Melee.

jobrien97:
- Characters now scream on frame 1 of their knockback statuses, rather than frame 6
- Knockback voice clips while under 100% now play 1/3 as frequently as before

Resolves #819


jobrien97:
To whomstever is reviewing this:
The `StopModule::is_stop` check and frame 1.1 call are there to prevent voice clips from playing during hitlag. For instance, there are cases where DamageFlyHi’s first frame will occur during hitlag, and sometimes not. The frame 1.1 call will ensure a voice clip will play even if the animation starts during hitlag. And if it doesn’t, a knockback voice clip will not play if one is already playing, so a clip will simply not play on frame 1.1 if one does on frame 1.0.

WuBoytH:
So _this_ is your solution to the hitlag conundrum
I... Probably would have just slapped the voices on frame 2, but if it works it works.
Lgtm

jobrien97:
> So _this_ is your solution to the hitlag conundrum I... Probably would have just slapped the voices on frame 2, but if it works it works. Lgtm

Frame 2 is still pretty delayed at higher percents. There seems to be a motion rate applied during kb

WuBoytH (dismissed):
LGTM

PR #834, opened 2022-07-18
Reinstate wavedash item pickups

### LINKED ISSUE #781: Item pickup with wavedash broken on latest nightly
it seems in the recent nightly builds with the updated ecb changes that it is no longer possible to pick up items with a grounded wavedash. 

It seems the only way to do this is to intentionally delay my wavedash until I have been airborne for 2 frames.

I understand if the smoother wavedash takes priority over a niche mechanic like this. 

But it would be amazing if you could keep it in somehow.

-Analog <3

jobrien97:
Perfect wavedashes now run `sub_escape_air_common()`
Resolves #781


WuBoytH (approved):
LGTM

PR #836, opened 2022-07-18
Dr. Mario - remove multihit upB variant

jobrien97:
Changing hitbox properties based on hitting the initial sweetspot or not was unnecessary bloat.

UpB is now a single hit sweetspot + weak late hit, like before. The combo hitbox for grounded upB cancel is still there though.


WuBoytH (approved):
lgtm

PR #839, opened 2022-07-18
Robin - slight adjustments

### LINKED ISSUE #224: Robin cannot jump out of neutral b if holding down.
When charging neutral b (thunder), Robin is able to jump out of the charge, but not while holding down until you press in another direction. I assume this is caused by float being added as you can in vanilla, but either way it makes floating out of a neutral b charge to be unnecessarily precise in timing or else you get stuck in charge.

jobrien97:
- Float now consumes 1 bar of Elwind every 15 frames, rather than 2 bars every 30 frames.
- Robin can now jump cancel out of aerial Thunder charge when left stick is held down (previously they were the only character with a chargeable move that couldn't do so)

Resolves #224


WuBoytH (approved):
LGTM

PR #842, opened 2022-07-19
Ryu - fix airdash momentum out of hitstun

jobrien97:
Airdash will no longer go zooooooom if used out of hitstun


WuBoytH (approved):
lgtm

PR #849, opened 2022-07-20
Chrom/Lucina - add aerial sideB1 hop

### LINKED ISSUE #332: Chrom Side B Momentum
Chrom side B no longer has the slight float capabilities from last patch. This was discussed in the Chrom channel and calc mentioned it needing a rework. Just adding the issue to make sure it isn't forgotten.

jobrien97:
Chrom and Lucina have been given the same sideB1 momentum logic as Roy/Marth

Resolves #332

### DIFF
--- a/fighters/chrom/src/lib.rs
+++ b/fighters/chrom/src/lib.rs
@@ -4,7 +4,7 @@
 
 pub mod acmd;
 
-//pub mod status;
+pub mod status;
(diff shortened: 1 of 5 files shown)


WuBoytH (approved):
USING THE INIT STATUS LET'S GEEEEEEET THIS MONEEEEEEY

PR #855, opened 2022-07-21
Fix Explosion VFX Size.

HisashiASM:
Nothing special :), have a nice day.


Riah321:
LGTM. Thank you :)

Riah321:
LGTM. Thank you :)

WuBoytH:
As far as I understand, something like this won't... *build*, right? You'd need something in the build script that will rename the .lua to a .lc for when the file gets packaged?

WuBoytH:
Oh right uh move this to hdr-private.

PR #870, opened 2022-07-23
Falco - fix Shine stall

### LINKED ISSUE #862: Falco is still not recovering from his car accident even after the supposed fixes in the new nightly.
Falco still seems to fall out of shine and can't shine stall even with the fixes in the nightly. Jump cancelling the shine is also still affected by this.

HisashiASM:
We need to keep these momentum changes to falco until we have a better reason to remove them

Fixes #862


yoonikoo:
Feels amazing. The Bird has recovered from his unfortunate car accident that shattered both of his kneecaps. LGTM

WuBoytH (approved):

PR #887, opened 2022-07-29
Boost Running bugfix

### LINKED ISSUE #875: Mewtwo Run Turnaround Bug
Mewtwo gets a burst of speed during the beginning of their run turnaround animation. The increased speed lasts for about a second and then their speed decreases back to their normal run speed.

jobrien97:
Fixes an issue where some characters always had Boost Running available, even when run turn was not performed out of a Sticky Walk.

Fixes #875

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -209,6 +209,8 @@ pub mod vars {
 
             pub const DISABLE_BACKDASH: i32 = 0x1000;
             pub const IS_MOONWALK: i32 = 0x1001;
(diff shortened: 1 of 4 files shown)


jobrien97:
There is quite literally no other way to detect a character's dash-to-run frame programmatically, as `is_enable_transition_term(FIGHTER_STATUS_TRANSITION_TERM_ID_DASH_TO_RUN)` only returns true for a single frame *if* the stick is held forward. Ideally we go through every single character's Dash game ACMD and store their dash-to-run frame in ParamModule, but that day is not today.

WuBoytH (approved):
LGTM

PR #898, opened 2022-08-04
Mr. Game & Watch - dthrow adjustments

### LINKED ISSUE #889: Funny G&W down throw is a little too funny
Dthrow at ledge when G&W is facing into the stage forces you to walltech or die instantly. To recreate it roll backwards into the ledge as G&W, then grab and down throw your opponent. As a hilarious joke they will have to tech or get Mufasa'd.

jobrien97:
- Throw release point reverted to in front of G&W
- If spiked offstage with dthrow, you are now able to cancel hitstun after 8 frames (pseudo-meteor-cancel)

Resolves #889


WuBoytH (approved):

Riah321:
Tested in game, looks good to me. the grab at ledge is an interesting callout, though it can't be teched. consider moving the release point farther into his body so that techrolling on either side is equally viable.

PR #899, opened 2022-08-04
Turnaround special/B-reverse reimplementation

### LINKED ISSUE #890: Little Mac can B-Reverse his Neutral Special and Up Special twice.
https://streamable.com/xvlctj

jobrien97:
Turnaround special/B-reverse now uses vanilla logic, rather than our hacky workaround method.

In other words, inputs for turnaround special/B-reverses should now be identical for moves that were not reversible in vanilla.

Should remove the jank with our custom B-reversible moves.

Fixes #890

### DIFF
--- a/fighters/brave/src/opff.rs
+++ b/fighters/brave/src/opff.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-utils::import_noreturn!(common::opff::{fighter_common_opff, check_b_reverse});
+utils::import_noreturn!(common::opff::fighter_common_opff);
 
 unsafe fn nspecial_cancels(fighter: &mut L2CFighterCommon) {
     //PM-like neutral-b canceling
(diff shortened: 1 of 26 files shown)


WuBoytH (approved):
This will remove "tech" born from the old special-reverse implementation, like Falcon's "Wavekick". Expect backlash (which I completely approve of).

WuBoytH (changes requested):
Ah sorry I need to make an addendum
since the status_attr field is a bitmask, you need to make sure that
1. You add the value of *FIGHTER_STATUS_ATTR_START_TURN to the bitmask, instead of setting the bitmask itself.
2. You check if the bitmask *already contains* FIGHTER_STATUS_ATTR_START_TURN before adding to it.

WuBoytH (approved):
Free to merge once someone else tests it in-game

MadVillainLordQ:
Tested in game, LGTM

PR #903, opened 2022-08-04
Tumble Threshold Adjustment

MadVillainLordQ:
- Corrected an oversight in which tumble threshold was reached at 96 units of kb instead of 90 units.


Riah321:
Tested and works in game. Looks good to me.
~~Please ship it right now I'm begging you throws are so jank I miss my techchases I'm screaming and crying rn~~

WuBoytH (approved):

PR #905, opened 2022-08-04
Fix wavelands after aerial attacks

jobrien97:
Fixes an issue where wavelands would only have 4f of landing lag if aerial attack's endlag was interrupted with an airdodge.


WuBoytH (approved):
Code looks functionally soild, test in game before merging

PR #906, opened 2022-08-04
Remove random tripping

jobrien97:
Removed the mechanic that allowed any move to have a random trip chance.

Tripping has been changed into an all-or-nothing mechanic; either a move trips 100% of the time, or doesn't trip at all.

Currently, the only tripping moves that remain are Kazuya's Stature Smash, Pichu's ftilt, and Diddy's banana.


WuBoytH (dismissed):
I didn't know this was a thing but now I know and knowing is half the battle GI Joe
Looks solid, not sure if people will like the change but hey that's not me to decide

Snakkattakk:
But why though? Some of the identity of moves like Kirby down tilt is they're ok on hit, but if they trip and you can react and confirm off of it you get big punishes. I feel like this is kind of just removing something that doesn't need to be.

MadVillainLordQ:
Tried in game, LGTM.

PR #907, opened 2022-08-04
Suicide throw adjustments

### LINKED ISSUE #901: Kirby and Meta Knight up throw allow for space travel.
Both Meta Knight and Kirby's up throw can be mashed out of at the height of ninja drop. Meta knight has slightly more humorous results. On the bright side, this bug is actually canonically accurate!
I'm assuming this is because they share code and MK went untested. Considering that you can't actually kirbycide with this move anyways I'm requesting that mashing out of it be removed.

jobrien97:
Suicide throws (Kirby fthrow/bthrow, ROB uthrow) are now only able to be mashed out of during their descent, before the thrower reaches the ground. This makes them more reliable when used on-stage for combos.
Mash-out difficulty still scales based on % difference between thrown player and thrower.

Fixes #901

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -216,6 +216,8 @@ pub mod vars {
 
             pub const IS_JAB_LOCK_ROLL: i32 = 0x1000;
 
+            pub const SUICIDE_THROW_CAN_CLATTER: i32 = 0x1000;
+
             // ints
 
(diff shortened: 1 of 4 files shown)


WuBoytH (approved):

Riah321:
Tested in game. LGTM, though mashing and throwing by myself wasn't easy.

MadVillainLordQ:
Tried in game, LGTM.

PR #915, opened 2022-08-08
Resolve Tilt button not triggering AB Smash

techyCoder81:
Currently, tilt button cannot be used to trigger AB smash. This is because we clear all smash inputs if a TiltAttack input is found.

Instead, this change makes it so that we only clear smash inputs if the player has not pressed an *explicit* smash button. For example, smash stick and AB Smash are both internally register as a Smash button. However, a manually input smash attack using a regular Attack button does not register a Smash button prior to cat flag mappings. Therefore this operation should be safe.

Please test this PR to ensure that the change works as expected.

### DIFF
--- a/utils/src/modules/input.rs
+++ b/utils/src/modules/input.rs
@@ -473,7 +473,7 @@ fn exec_internal(input_module: &mut InputModule, control_module: u64, call_origi
         *(control_module as *mut *mut BattleObjectModuleAccessor).add(1)
     };
 
-    if triggered_buttons.intersects(Buttons::TiltAttack) {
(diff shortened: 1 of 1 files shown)


WuBoytH (dismissed):

MadVillainLordQ:
Unfortunately, this was not able to resolve the issue. Pressing A+B with tilt button enabled results in a special attack.

jobrien97 (changes requested):
does not solve the issue

Riah321:
It did not, in fact, fix the issue.

PR #919, opened 2022-08-09
Fix JC usmash logic

### LINKED ISSUE #918: Reimplement Jump Cancel Up Smash
In Vanilla, if you enter jumpsquat, release the jump button, and press A while holding Up, you will *always* get an Up Smash. This will be updated so that it will instead only work if the stick has been flicked upwards, instead of needing to be tilted up at all.

WuBoytH:
In Vanilla, if you have up held on the left stick for any amount of time, enter jumpsquat, let go of jump and press A, you will always get usmash.
Now you must flick the stick to get a JC usmash. (this makes stuff like buffered uairs easier)

Resolves #918

### DIFF
--- a/fighters/common/src/general_statuses/jumpsquat.rs
+++ b/fighters/common/src/general_statuses/jumpsquat.rs
@@ -135,16 +135,7 @@ unsafe fn status_JumpSquat_Main(fighter: &mut L2CFighterCommon) -> L2CValue {
         );
         return 0.into();
     }
-    // Walks through and clears all item throw commands.
-    if fighter.global_table[CMD_CAT3].get_i32() & *FIGHTER_PAD_CMD_CAT3_FLAG_ITEM_LIGHT_THROW_ALL != 0 {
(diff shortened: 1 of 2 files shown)


Riah321:
JC item toss logic should be as follows

Option 1: (preferred for ease of use) Remove JC item toss, make any toss during JS buffer the air version
Option 2: (preferred for depth) Any item toss input while holding jump during JS will buffer the air version. Item toss after releasing jump will input the grounded toss varient. This allows for buffered full hop _and_ short hop item tosses, as well as JC tosses

Otherwise looks solid to me :)

jobrien97:
In Vanilla, if you have up held on the left stick for any amount of time, enter jumpsquat, let go of jump and press A, you will always get usmash.
Now you must flick the stick to get a JC usmash. (this makes stuff like buffered uairs easier)

techyCoder81 (approved):
looks pretty simple to me, just should be tested

jobrien97:
Tested ingame, works

PR #924, opened 2022-08-14
Aesthetic wavedash improvements

### LINKED ISSUE #916: Adjust jumpsquat/landing animations
Certain characters could use slight adjustments to their jumpsquat/landing animations to make wavedashing/wavelanding feel smoother.

A single frame will be shaved off of the beginning of landing animations globally.

Character jumpsquats to adjust (so far):
- Samus
- Fox
- Pikachu
- Luigi
- Ness
- Falco
- ICs
- Pichu
- Young Link
- Squirtle
- Diddy Kong
- Lucas
- Wolf
- Little Mac
- Duck Hunt
- Ryu
- Ken
- Cloud
- Simon
- Richter
- Terry
- Sephiroth

jobrien97:
In Ultimate, characters' animation philosophies for jumpsquat/landing animations are quite different than past games. Several characters appear to almost leave the ground on the last frame of their jumpsquat animation, or will be near-airborne on the first frame of their landing animations.
This does not translate well into our engine, with proper wavedashes/wavelands in place; it results in wonky-looking movement.

This change does two things:
1. Shaves off the **first** frame from *all* characters' heavy landing animations
2. Shaves off the **last** frame from *some* characters' jumpsquat animations.

This results in much smoother-looking wavedashes/wavelands, akin to Melee.

Old:
https://user-images.githubusercontent.com/47401664/184538481-b3239ec6-999b-4ef2-9add-c15693fdaece.mp4

New (more noticeable ingame at 60fps):
https://user-images.githubusercontent.com/47401664/184538334-da5ac9cf-0e33-4ba2-88fb-c19c83cee7d6.mp4

Resolves #916

### DIFF
--- a/fighters/common/src/function_hooks/get_param.rs
+++ b/fighters/common/src/function_hooks/get_param.rs
@@ -77,6 +77,10 @@ pub unsafe fn get_param_float_hook(x0 /*boma*/: u64, x1 /*param_type*/: u64, x2 
     // For articles
     let owner_module_accessor = &mut *sv_battle_object::module_accessor((WorkModule::get_int(boma, *WEAPON_INSTANCE_WORK_ID_INT_LINK_OWNER)) as u32);
 
+    if x1 == hash40("landing_frame") {
+        return original!()(x0, hash40("landing_frame"), 0) + 1.0;
+    }
+    
     if x1 == hash40("param_trenchmortarbullet") && x2 == hash40("speed_x") {
 		if fighter_kind == *WEAPON_KIND_SNAKE_TRENCHMORTAR_BULLET {
 			return ControlModule::get_stick_x(boma) / 1.5 * PostureModule::lr(boma);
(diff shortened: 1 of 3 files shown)


fighters/common/src/function_hooks/get_param.rs
```diff
@@ -77,6 +77,12 @@ pub unsafe fn get_param_float_hook(x0 /*boma*/: u64, x1 /*param_type*/: u64, x2
     // For articles
     let owner_module_accessor = &mut *sv_battle_object::module_accessor((WorkModule::get_int(boma, *WEAPON_INSTANCE_WORK_ID_INT_LINK_OWNER)) as u32);
 
+    // Coupled with "landing_heavy" change in change_motion hook
```
WuBoytH, line 80: This should be handled in params instead

WuBoytH (changes requested):

fighters/common/src/function_hooks/get_param.rs
jobrien97: I didn't want to change this in params, as it would be confusing to someone without context why heavy landing lag was 1 frame less ingame than defined in params. It would also be easy to forget the addition of an extra frame if heavy landing lag is adjusted for a character in params.
This way the change can be easily altered/reverted if need be, as well.

WuBoytH: I see, my "param hook bad" mentality kicked in real quick on that one
In that case sure I'll allow it

WuBoytH (dismissed):

WuBoytH (approved):

PR #925, opened 2022-08-14
Stick Sensitivity Adjustments

MadVillainLordQ:
Corrects behavior on low and high stick sensitivities:

Current Stick Sensitivity is as follows:

Low: Smash attack can occur until frame 5 of analog press
Normal: Smash attack can occur until frame 4 of analog press
High: Smash attack can occur until frame 7 of analog press

The new setting will be as follows:
Low: Smash attack can occur until frame 2 of analog press
Normal: Smash attack can occur until frame 4 of analog press
High: Smash attack can occur until frame 6 of analog press

### DIFF
--- a/romfs/source/fighter/common/param/common.prcxml
+++ b/romfs/source/fighter/common/param/common.prcxml
@@ -3,6 +3,8 @@
   <int hash="precede">5</int>
   <int hash="precede_extension">0</int>
   <float hash="dash_s4_frame">4</float>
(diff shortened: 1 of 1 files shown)


jobrien97 (approved):
lgtm

WuBoytH:
For the record, in case the original values look strange to you, vanilla's stick sensitivity is 6, HDR changed it to 4 at some point without changing the other options

PR #931, opened 2022-08-16
Drift DI rework

jobrien97:
- DDI now uses your *initial* horizontal knockback velocity (IHKBV) to calculate your DDI value, rather than recalculating every frame throughout the entirety of hitstun.
- DDI now scales linearly based on your IHKBV
   - The higher your IHKBV, the lower your DDI strength
- DDI is now **disabled** once you are hit with high enough IHKBV
  - Meaning DDI should no longer impact survivability

The nitty gritty:
- Base DDI strength is 0.0075 units/frame, at 0.0 IHKBV
- This decreases linearly until 3.0 IHKBV
- At 3.0 IHKBV or more, your DDI strength is 0.0

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -225,6 +225,11 @@ pub mod vars {
 
             pub const DOWN_STAND_FB_KIND: i32 = 0x1000;
 
+            // floats
+
+            pub const INITIAL_KNOCKBACK_VEL_X: i32 = 0x1000;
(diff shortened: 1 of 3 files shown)


O1yyy:
does it feel gooder

SkewedAskew (approved):
Tested with all notched angles with no/right/left DDI. All three in every case exited hitstun in the exact same position.

In other words, lgtm

Riah321:
It does, in fact, feel more gooder.

PR #934, opened 2022-08-17
Reimpl'd Illusion

WuBoytH:


WuBoytH:
There's currently a weird thing where shortening Illusion in the air will not make you have any air friction. Is this intentional?

WuBoytH:
> There's currently a weird thing where shortening Illusion in the air will not make you have any air friction. Is this intentional?

This has been fixed, the game overrides your air brake with a much smaller value, letting you drift further than normal.

TeamPearson:
Tested this out on all three characters timing feels good.

jobrien97 (approved):
lgtm, tested ingame as well

PR #939, opened 2022-08-18
Chrom dair hitbox ID Fix

MadVillainLordQ:
Fixed an error on dair's hitbox ID where sweetspot ground hitbox had same hitbox ID as aerial sweetspot hitbox


jobrien97 (approved):
good catch, lgtm

PR #944, opened 2022-08-18
Add dash dance SFX

### LINKED ISSUE #797: [Audio request] Make initial dash turnarounds have a sound
Dash dancing in Melee/PM had a sound for each time you turned around in an initial dash, here it's just silence after the first despite the huge amount of visual smoke fx lmao

jobrien97:
Previously, no SFX would play if you input a backdash during dash. Only the initial dash would have sound, which made for weirdly silent dash dancing.

Removed the `SET_PLAY_INHIVIT` calls from dash's sound ACMD, which disallowed dash SFX from replaying until 20 frames had passed since the previous sound.

Additionally, dash SFX volume has been reduced by 50% across the board.

Resolves #797


WuBoytH (dismissed):
Jesus that's a lot of script

WuBoytH (approved):

PR #949, opened 2022-08-20
Revert Home Run Bat SFX

WuBoytH:
Before, Ness's FSmash's strong hit played the Home Run Bat SFX. Now, it plays Ness's usual Bat SFX.

In addition, the weaker hits that used to play Ness's Bat SFX now play Medium Punch, to differentiate them from the strong hit.


jobrien97:
not a change we really wanted for now

Issue #965, opened 2022-08-24
Jumping out of shield does not require a new shield input for Z-aerials


Riah321:
This means Z-air characters do not get to comfortably aerial out of shield.

moklmaru:
Being tracked by #2115

Issue #444, opened 2022-04-05
Zair behavior request


Scorpio3002:
Can we possibly get Zair characters working like in PM? Currently, zair will trigger any time the Shield button is being held. PM ignores any "old" Shield input, so you can input an aerial out of shield or during a shield-drop without first releasing the trigger.

Riah321:
~~This has been addressed in the new beta. Please close this issue :)~~
I lied. I hope this gets fixed soon

moklmaru:
being tracked by #2115

PR #970, opened 2022-08-24
Shield adjustments

### LINKED ISSUE #724: Shield + Taunt Input buffers shield drop
Shield plus taunt input can automatically buffer shield drop. The implementation I'd suggest is that shield plus taunt works similarly to other shield "lock" options such as shield + B and shield + shield. This has been discussed previously among devs, and this post is reminder for techy as has mentioned he will take care of it.

### LINKED ISSUE #775: Add Melee/PM UCF Shielddropping
current methods of shielddropping are good but itd be good for cross-game playing to be able to use the same methods as other games. 

Currently shielddropping feels clunky having to use multiple shield buttons etc compared to just a quick smooth stick flick. 

UCF shielddropping follows the following mechanics:
Stick position at the rim and in a state such that you cant perform a roll with it, then the spotdodge stick range will move much lower so the range for shielddrop is much more accessible. 

Exact numbers used in PM shielddrop since i can give them:
- tracking at rim: 
 This is all basically just a check for "holding stick close to rim and has been holding with a consistent X component for 4 frames to disable roll"

> Stick X position must be non-zero
> Stick X direction must be same as previous frame for 4 frames
> if (|x|+0.125)^2 + (|y|+0.125)^2>1.0 then enable UCF shielddrop ranges
- shielddrop ranges:
 values used specifically in P+
> change spotdodge y req from -0.75 to -0.8
> change shielddrop y req from -0.71 to -0.675

jobrien97:
- Spotdodge/platdrop stick thresholds changed to match Melee/PM, to widen the base range for shielddropping:
   - Spotdodge stick y threshold: -0.7 -> -0.75
   - Platdrop stick y threshold: -0.6875 -> -0.71
- Axe Dropping works again
- Melee/PM UCF shielddropping added
   - If your shield is angled using the Axe method, the platdrop stick threshold will move higher, and the spotdodge stick threshold will move much lower, so the range for a shielddrop is much more accessible
   - The thresholds shift as follows:
      - Spotdodge stick y threshold: -0.75 -> -0.8
      - Platdrop stick y threshold: -0.71 -> -0.675
- Shielddrops can no longer be buffered with a taunt input
- Parry reflect projectile speed multiplier increased 0.9 -> 1.0 (parrying a projectile now sends them back at the same speed at which they came)

Resolves #724 
Resolves #775


Riah321:
Please consider also addressing issues #444 and #965

fighters/common/src/general_statuses/shield/misc.rs
```diff
@@ -161,9 +166,22 @@ pub unsafe fn sub_guard_cont(fighter: &mut L2CFighterCommon) -> L2CValue {
...
+            // If your left stick is near the rim and you haven't triggered a roll
+            let escape_fb_stick_x = WorkModule::get_param_float(fighter.module_accessor, hash40("common"), hash40("escape_fb_stick_x"));
+            if fighter.global_table[STICK_X].get_f32().abs() > escape_fb_stick_x
+            && cat2 & *FIGHTER_PAD_CMD_CAT2_FLAG_STICK_ESCAPE_F == 0
+            && cat2 & *FIGHTER_PAD_CMD_CAT2_FLAG_STICK_ESCAPE_B == 0
+            && !VarModule::is_flag(fighter.battle_object, vars::common::status::ENABLE_UCF) {
+                // Enable UCF shielddrop thresholds
+                // change spotdodge y req from -0.75 to -0.8
+                // change platdrop y req from -0.71 to -0.675
+                VarModule::on_flag(fighter.battle_object, vars::common::status::ENABLE_UCF);
```
WuBoytH, line 179: If you enable UCF, I assume on the next frame it'll use the UCF drop threshold?

jobrien97: > If you enable UCF, I assume on the next frame it'll use the UCF drop threshold?

Technically should be available starting on the same frame, but don't think it's possible to input both the shield tilt and shielddrop on the same frame lol

WuBoytH (dismissed):

WuBoytH (approved):

PR #975, opened 2022-08-25
Glide Toss rework

### LINKED ISSUE #953: FGC backwards glide toss
All four characters with auto-turn get reversed momentum if inputting a glide toss on frame 4 or later of roll

https://user-images.githubusercontent.com/55216313/185782162-83afcf3e-ec5a-4814-86f5-8a267fa4ecf3.mp4

jobrien97:
Glide Tossing no longer applies artificial speed to your character when canceling roll into an item throw. Now, your roll's animation speed on the last frame before being canceled is carried over into your item throw animation. Due to characters' varying roll animations, this means the timing for a "good" Glide Toss will vary for each character & roll direction. Generally, the later into the roll you input a Glide Toss, the further you'll slide.

Additionally, the grounded Glide Toss window has been made consistent with the Aerial Glide Toss window, at 5 frames.

Also fixes an issue where the autoturn characters would sometimes get boosted in the wrong direction when Glide Tossing.

Fixes #953

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -189,7 +189,7 @@ pub mod vars {
             pub const RAR_LENIENCY: i32 = 0x0006; // Only ever gets set, goes effectively unused.
             pub const CURRENT_MOMENTUM_SPECIALS: i32 = 0x0007;
             pub const DOUBLE_JUMP_TIMER: i32 = 0x0008; // Only used by Lucas, and it's commented out, goes unused.
-            pub const ROLL_DIR: i32 = 0x0009;
+            pub const ROLL_SPEED: i32 = 0x0009;
             pub const LEDGE_POS: i32 = 0x000A;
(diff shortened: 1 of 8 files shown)


WuBoytH (commented):
Correct me if I'm wrong, but are you sure you need to store the speed from your roll? I believe the kinetic type Item Throw uses (`FIGHTER_KINETIC_TYPE_UNIQ`) supports carrying over leftover kinetic motion.

jobrien97:
> Correct me if I'm wrong, but are you sure you need to store the speed from your roll? I believe the kinetic type Item Throw uses (`FIGHTER_KINETIC_TYPE_UNIQ`) supports carrying over leftover kinetic motion.

Nah we’ve tried simply changing status, and it doesn’t carry any momentum over.

WuBoytH:
`sub_status_end_EscapeFB`.

If you know you know.

WuBoytH (approved):

PR #981, opened 2022-08-28
Tether ledge occupancy bugfixes

### LINKED ISSUE #848: Grabbing ledge while tethered but without reeling in trumps opponents.
If you're fixing this, please make sure to keep the ability to grab ledge without reeling in! It's very fun and makes tethers more interesting.

jobrien97:
Fixed up logic for tether recoveries’ ledge occupancy.

You should no longer ever be trumped by a tethering opponent.

Fixes #848


jobrien97:
Yes, I realize housing the logic in change_status is undesirable, it will be implemented within the respective statuses eventually

WuBoytH (approved):
JOB to ledges:
image

PR #993, opened 2022-08-30
Tether Trump DI removal

chrispo-git:
Tethers no longer have Directional Influence affect them when trumped


Riah321:
This isn't going to be implemented, right?

jobrien97:
Not being used

PR #996, opened 2022-09-01
Compatibility for Skyline update

jobrien97:
Replace dependencies and instruction patches

### DIFF
--- a/fighters/common/src/function_hooks/mod.rs
+++ b/fighters/common/src/function_hooks/mod.rs
@@ -42,9 +42,9 @@ pub fn install() {
     unsafe {
         // Handles getting rid of the kill zoom
         const NOP: u32 = 0xD503201Fu32;
-        skyline::patching::patch_data(utils::offsets::kill_zoom_regular(), &NOP);
-        skyline::patching::patch_data(utils::offsets::kill_zoom_throw(), &NOP);
+        skyline::patching::Patch::in_text(utils::offsets::kill_zoom_regular()).nop();
+        skyline::patching::Patch::in_text(utils::offsets::kill_zoom_throw()).data(NOP);
         // Changes full hops to calculate vertical velocity identically to short hops
-        skyline::patching::patch_data(0x6d2188, &0x52800015u32);
+        skyline::patching::Patch::in_text(0x6d2188).data(0x52800015u32);        
     }
 }
--- a/fighters/common/src/function_hooks/mod.rs
+++ b/fighters/common/src/function_hooks/mod.rs
@@ -41,9 +41,9 @@ pub fn install() {
 
     unsafe {
         // Handles getting rid of the kill zoom
-        const NOP: u32 = 0xD503201Fu32;
+        const KILL_ZOOM_DATA: u32 = 0xD503201Fu32;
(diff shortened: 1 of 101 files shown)


fighters/common/src/function_hooks/mod.rs
```diff
@@ -42,9 +42,9 @@ pub fn install() {
     unsafe {
         // Handles getting rid of the kill zoom
         const NOP: u32 = 0xD503201Fu32;
-        skyline::patching::patch_data(utils::offsets::kill_zoom_regular(), &NOP);
-        skyline::patching::patch_data(utils::offsets::kill_zoom_throw(), &NOP);
+        skyline::patching::Patch::in_text(utils::offsets::kill_zoom_regular()).nop();
+        skyline::patching::Patch::in_text(utils::offsets::kill_zoom_throw()).data(NOP);
```
techyCoder81: is there a reason one of these uses `.nop()` and the other uses `.data(NOP)`

plugin/Cargo.toml
```diff
@@ -29,8 +29,8 @@ webpki = { git = "https://github.com/skyline-rs/webpki" }
 getrandom = { git = "https://github.com/skyline-rs/getrandom" }
 
 
-[patch.'https://github.com/skyline-rs/skyline-web']
-skyline-web = { git = "https://github.com/blu-dev/skyline-web", branch = "less-panics" }
```
techyCoder81, line 33: what is the reasoning for this change?

techyCoder81 (commented):
Has the build produced by this been somewhat tested to make sure it mostly works as expected? Especially, the mechanics who used the old patching api, to make sure they still work like they did before?

fighters/common/src/function_hooks/mod.rs
jobrien97: > is there a reason one of these uses `.nop()` and the other uses `.data(NOP)`

The NOP in `.data(NOP)` is actually a reference to an address, defined a few lines above. I’m not sure if it functionally does the same thing as .nop(), but that’s how Blujay had it implemented.

jobrien97:
> Has the build produced by this been somewhat tested to make sure it mostly works as expected? Especially, the mechanics who used the old patching api, to make sure they still work like they did before?

Yes, been tested for several weeks with no crashes/bugs regarding what those patches address.

plugin/Cargo.toml
jobrien97: > what is the reasoning for this change?

Blujay needs to update his fork to resolve build errors lol, which he won’t have time for for who knows how long. I assume this is used for the game modes page?

fighters/common/src/function_hooks/mod.rs
techyCoder81: it might be wise to rename that variable to REPLACEMENT_DATA or something if we want to leave it unknown. Right now, this is the kind of thing someone might want to just fix in passing... and that could be bad.

plugin/Cargo.toml
techyCoder81: primarily, yes. I'm not sure what all "less panics" does besides... have less panics lol. We may want to look at the diff between the two branches to see what it entails.

PR #1004, opened 2022-09-06
check for fighter or weapon

### LINKED ISSUE #1003: Fix param hooks so that it checks the boma's battle object category first
Because Fighter IDs and Weapon IDs can overlap, it's possible that param hooks can return undesirable results if the category isn't checked first.

WuBoytH:
resolves #1003


jobrien97 (dismissed):
lgtm

jobrien97 (approved):

PR #1012, opened 2022-09-11
Prevent grab input from triggering attack/shield on ground

### LINKED ISSUE #1011: Disable grab button from triggering attack/shield
Grab button should only cause grab

jobrien97:
Grab button can no longer cause an attack or shield on the ground

Resolves #1011

### DIFF
--- a/fighters/common/src/general_statuses/mod.rs
+++ b/fighters/common/src/general_statuses/mod.rs
@@ -279,7 +279,7 @@ unsafe fn sub_transition_group_check_air_escape(fighter: &mut L2CFighterCommon) 
 
 #[skyline::hook(replace = L2CFighterCommon_sub_transition_group_check_ground_attack)]
 unsafe fn sub_transition_group_check_ground_attack(fighter: &mut L2CFighterCommon) -> L2CValue {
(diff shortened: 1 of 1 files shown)


WuBoytH (dismissed):
This will not be enough to prevent Grab from triggering these actions, especially in the case of Shielding, where the only check is whether or not the guard button is pressed, ignoring all CAT flags.

jobrien97:
> This will not be enough to prevent Grab from triggering these actions, especially in the case of Shielding, where the only check is whether or not the guard button is pressed, ignoring all CAT flags.

uhhh I exit the shield transition func if Catch cmd cat is active, so that shouldn't matter (?)

WuBoytH:
> > This will not be enough to prevent Grab from triggering these actions, especially in the case of Shielding, where the only check is whether or not the guard button is pressed, ignoring all CAT flags.
> 
> uhhh I exit the shield transition func if Catch cmd cat is active, so that shouldn't matter (?)

You'll love this
holding Grab counts as holding Shield, doesn't matter if the cat flag is active
(this comment was written before reading the (original) PR description)

WuBoytH (approved):
Gameing

PR #1017, opened 2022-09-12
Perfect Pivot window adjustment

Riah321:
After talking with several community members and listening to discourse about melee/perfect pivots, I have come to the conclusion that a 3 frame window is the lowest we can go while having perfect pivots remain consistent and accessible. Perfect pivots have been a point of contention for some time because they get in the way of melee pivots. I believe this change will serve to help the two mechanics co-exist.


jobrien97 (approved):
needs more comments /s

PR #1019, opened 2022-09-12
Add support for changing the cliff hangdata at runtime

### LINKED ISSUE #1018: Add support for arbitrary ledge-grab boxes
Currently there is no way to change ledge-grab boxes outside of a select few based on the character.

blu-dev:
Resolves #1018

The game implementation appears to reset the hangdata at status change, so no need to worry about that

### DIFF
--- a/dynamic/src/ext.rs
+++ b/dynamic/src/ext.rs
@@ -468,6 +468,9 @@ pub trait BomaExt {
     // tech/general subroutine
     unsafe fn handle_waveland(&mut self, require_airdodge: bool) -> bool;
     unsafe fn shift_ecb_on_landing(&mut self);
+    unsafe fn set_front_cliff_hangdata(&mut self, x: f32, y: f32);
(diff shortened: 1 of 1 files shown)


jobrien97:
I think it might be more intuitive to have a single `set_cliff_hangdata` ext, that takes p1_x, p2_x, p1_y, p2_y as args which can be pulled from a ParamModule hangdata struct set up identically to vanilla hangdata, if that is feasible.
Defining 4 coordinates for the entire box and letting the ext handle the calculations in the background to create center/front/back points is much easier to understand imo

techyCoder81:
I would like to see someone (@jobrien97 ?) test this functionality in nightly as-is personally. There is no reason that this cannot be expanded upon *once its in*. ParamModule things can be easily integrated with this, but this PR exists purely to expose functionality which previously did not exist at all (which is hype). Recommend future PRs by users of this API to add more clean infrastructure as desired (including new boma exts, etc).

PR #1025, opened 2022-09-13
Fix tilt button causing shield drop

### LINKED ISSUE #1020: Tilt Button Causes Shield Drops
Holding shield and pressing the tilt button will cause you to shield drop, meaning you cannot use it to shield grab. Strangely, holding 2 shield buttons will allow you to shield grab with it.

jobrien97:
Fixes an issue where inputting tilt button while shielding would erroneously trigger a shield drop.

Fixes #1020


WuBoytH (approved):
the cat flags... what do they mean...

PR #1038, opened 2022-09-21
Revert B-Reverse/turnaround B stick threshold

jobrien97:
B-Reverse or turnaround B sensitivity 0.0 -> 0.3

A value of 0.0 did not take into account the stick's deadzones, and thus would trigger B-Reverse/turnaround B too easily.


MadVillainLordQ (approved):
LGTM

PR #1041, opened 2022-09-22
Duck Hunt Tweaks

elias-michaias:
All aerials:
- Removed intangibility of duck during aerials

Up-Air: 
- Removed FKB from first two hits
- Decreased KBG on first two hits
- Added BKB to first two hits

Down-Air:
- Adjusted grounded down-air angle to send diagonally behind Duck Hunt

Up-Special:
+ Able to cancel 3 frames earlier

Side-Special:
* Increased BKB and KBG
* Changed angle to 90 degrees (straight upward)
* Removed ability for opponents to clank with it, yielding in unreasonable frame advantage

Down-Special:
* Timer resets when training mode resets

Up-Tilt:
- Increased BKB and KBG
- Pushed launch angle even further back behind Duck Hunt by 4 degrees

Jab 1:
+ Inner hit pops upward when opponent is not knocked down
+ Slightly increased hitbox size
~ Locks opponents knocked down

Jab 2:
+ Inner hit pops upward when opponent is not knocked down
~ Locks opponents knocked down

Jab 3:
+ Changed launch angle to 70 degrees
+ Slightly improved FAF

Dash Attack:
+ Improved startup time and mobility 

GOAL:
Remove options that are excessively good without variation or require little to no thought; give better tools for playing the game and disable auto-loops that do not require reading or reacting to DI.


SkewedAskew:
Just a couple small things:

- Due to using FKB, the first two hits of uair are a constant +12 at max. Preferably this should be less at low percents. To adjust frame advantage, use BKB and KBG. Much like regular knockback, BKB controls advantage at 0 while KBG controls how fast it increases. You'll have to play with the values, but somewhere between 20-35 for both is the range you'll probably want
- Your dash attack motion rate should be moved from opff to its acmd script. Use FT_MOTION_RATE(fighter, <rate>); on the respective frame you want, in this case frame 0 and frame 9
- You may want to shrink fair/bair/uair's hitboxes a bit, but I can understand if you want them that way to better hit the can
- Attacking Clay Pigeon with a move that deals about 5% or more causes it to automatically go into its multihit as if Special was pressed. If that's intentional then it's fine, just something to note

Otherwise looks good to me

elias-michaias:
> Just a couple small things:
> 
> * Due to using FKB, the first two hits of uair are a constant +12 at max. Preferably this should be less at low percents. To adjust frame advantage, use BKB and KBG. Much like regular knockback, BKB controls advantage at 0 while KBG controls how fast it increases. You'll have to play with the values, but somewhere between 20-35 for both is the range you'll probably want
> * Your dash attack motion rate should be moved from opff to its acmd script. Use FT_MOTION_RATE(fighter, ); on the respective frame you want, in this case frame 0 and frame 9
> * You may want to shrink fair/bair/uair's hitboxes a bit, but I can understand if you want them that way to better hit the can
> * Attacking Clay Pigeon with a move that deals about 5% or more causes it to automatically go into its multihit as if Special was pressed. If that's intentional then it's fine, just something to note
> 
> Otherwise looks good to me

For the latter two: 
1. The compromise for these hitbox sizes previously amongst other balance team members was that either Duck Hunt's intangibility on aerials had to go, or the hitbox size had to get significantly reduced. Due to Duck Hunt's nature, I figured opting for the more vanilla route. Personally, I think it is very important to have these hitboxes, and is not overpowered/broken, but am willing to discuss more if the need arises or others feel that I am in error.
2. This is a feature since vanilla, if the clay pigeon is broken past it's damage threshold, it has a detonation. This is distinct from having pressed special and has a unique hitbox as well, ergo it ought to be retained. 

For the former two:
I will fix both of these promptly! Thanks so much for the review!

elias-michaias:
Finished requests. Ready for final review.

SkewedAskew (dismissed):
Lgtm

blu-dev:
needs more comments smh

jobrien97 (approved):
lgtm

PR #1051, opened 2022-09-29
Implement Dash Attack Slip-Offs and Roll-Offs

WuBoytH:
Implements Slip-Offs and Roll-Offs for Dash Attack.

Slip-Offs are when Dash Attack is fully canceled into the FALL status when you go over an edge.

Roll-Offs are when Dash Attack *continues the attack* after going over an edge. 

Examples of both are shown in the following video - Donkey Kong performs a Dash Attack Roll-Off while Young Link performs a Dash Attack Slip-Off. **These are *not* a part of the PR - they are here strictly for demonstration purposes.**

https://user-images.githubusercontent.com/26860994/193074167-8a6f1242-4d17-45f0-b7f6-030cf875bfe8.mp4

To enable Slip-Offs, run:
`VarModule::on_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_FALL);`

To enable Roll-Offs, run the Slip-Off command as well as:
`VarModule::on_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_CONTINUE);`

Both of these previous flags will stay on until you decide to turn them off, allowing you control over when in a Dash Attack you can slide off a ledge.

For the case of Roll-Offs, additional flags can be enabled to allow for air drift and enable the landing state.

The air drift flag will typically be put at the *end* of a Dash Attack, so you can drift when the attack is over and not be stuck in one place, falling to your death. Enable this by running:
`VarModule::on_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_DRIFT);`

The landing flag can be put anywhere, as long as you feel that you should be cancelling your Dash Attack on landing. The landing state used is the same as the one used for Auto-Cancels, so keep that in mind. Enable this by running:
`VarModule::on_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_LANDING);`

### DIFF
--- a/fighters/common/src/general_statuses/attackdash.rs
+++ b/fighters/common/src/general_statuses/attackdash.rs
@@ -162,7 +162,9 @@ unsafe extern "C" fn status_AttackDash_Main(fighter: &mut L2CFighterCommon) -> L
     // This is all behind an `is_situation_changed` check because we don't want this code running all of the time.
     if StatusModule::is_situation_changed(fighter.module_accessor) {
         if fighter.global_table[SITUATION_KIND].get_i32() != SITUATION_KIND_GROUND {
+            // Checks if your dash attack should roll off or not.
             if VarModule::is_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_CONTINUE) {
+                // Enables gravity.
                 sv_kinetic_energy!(
                     reset_energy,
                     fighter,
@@ -175,6 +177,7 @@ unsafe extern "C" fn status_AttackDash_Main(fighter: &mut L2CFighterCommon) -> L
                     0.0
                 );
                 KineticModule::enable_energy(fighter.module_accessor, *FIGHTER_KINETIC_ENERGY_ID_GRAVITY);
+                // Checks if you have an attack_air_dash motion_kind in your motion_list. If so, change to it.
                 if MotionModule::is_anim_resource(fighter.module_accessor, Hash40::new("attack_air_dash")) {
                     MotionModule::change_motion_inherit_frame(
                         fighter.module_accessor,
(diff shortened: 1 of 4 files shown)


fighters/common/src/general_statuses/attackdash.rs
```diff
@@ -131,26 +131,102 @@ unsafe extern "C" fn sub_attack_dash_uniq(fighter: &mut L2CFighterCommon, arg: L
...
+    && situation == *SITUATION_KIND_GROUND {
+        GroundModule::correct(fighter.module_accessor, GroundCorrectKind(*GROUND_CORRECT_KIND_GROUND));
+    }
+    // This block checks if you want to enable air drift. This code will only run once, and only while in the air,
+    // but it enables another flag that will be checked when your situation changes and renable the right kinetic type there.
+    if VarModule::is_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_DRIFT)
+    && situation != *SITUATION_KIND_GROUND {
+        VarModule::off_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_DRIFT);
+        VarModule::on_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_AIR_DRIFT_ENABLED);
+        KineticModule::change_kinetic(fighter.module_accessor, *FIGHTER_KINETIC_TYPE_FALL);
```
jobrien97, line 159: Wouldn't enabling ENERGY_ID_CONTROL suffice here? As well as line 194?

fighters/common/src/general_statuses/attackdash.rs
```diff
@@ -131,26 +131,102 @@ unsafe extern "C" fn sub_attack_dash_uniq(fighter: &mut L2CFighterCommon, arg: L
...
-        } else {
-            KineticModule::change_kinetic(fighter.module_accessor, *FIGHTER_KINETIC_TYPE_FALL);
-            VarModule::on_flag(fighter.battle_object, vars::common::ATTACK_DASH_SLIDEOFF);
+    // <HDR>
+    let situation = fighter.global_table[SITUATION_KIND].get_i32();
+    // This block is to force a ground correct type depending on if you enable sliding off or not.
+    if VarModule::is_flag(fighter.battle_object, vars::common::status::ATTACK_DASH_ENABLE_AIR_FALL)
+    && situation == *SITUATION_KIND_GROUND {
+        GroundModule::correct(fighter.module_accessor, GroundCorrectKind(*GROUND_CORRECT_KIND_GROUND));
+    }
```
jobrien97, line 152: Unsure of the purpose of this block, and why it needs to run every frame.

WuBoytH: the default correct kind is the one that doesn't allow slipoffs
GroundModule::correct changes the correct kind for the frame it's run, and we only want to change it when slip/roll-offs are enabled.

fighters/common/src/general_statuses/attackdash.rs
WuBoytH: it wouldn't work, because the last time Control is enabled is during DASH *because it's required to do dash attack), simply enabling control here would use the kinetic params for DASH. Or at least, that's my understanding of it anyway.
changing kinetic type to KINETICC_TYPE_FALL here is much simpler as it carries your MOTION momentum *and* uses the right params for air drift.

fighters/common/src/general_statuses/attackdash.rs
jobrien97: Isn't that what's being done in the pre status tho? Did you want to deprecate the usage of ParamModule?

WuBoytH: You seem to have missed it in my PR description.

> Both of these previous flags will stay on until you decide to turn them off, allowing you control over when in a Dash Attack you can slide off a ledge.

jobrien97: bold of you to assume I have reading comprehension....

jobrien97 (approved):
lgtm

PR #1073, opened 2022-10-02
Non-tumble hitstun landing adjustment

jobrien97:
- When landing from non-tumble knockback, hitstun is now halved **only** if you have successfully CC'd
- All other landing scenarios retain full hitstun

### DIFF
--- a/fighters/common/src/function_hooks/change_status.rs
+++ b/fighters/common/src/function_hooks/change_status.rs
@@ -94,16 +94,6 @@ unsafe fn change_status_request_from_script_hook(boma: &mut BattleObjectModuleAc
                 }
             }
         }
-        // Checks whether you have landed on the first frame of non-tumble knockback
-        // This is only possible by CCing
-        if boma.is_status(*FIGHTER_STATUS_KIND_DAMAGE_AIR) && next_status == *FIGHTER_STATUS_KIND_LANDING {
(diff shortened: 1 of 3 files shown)


fighters/common/src/function_hooks/change_status.rs
```diff
@@ -94,6 +94,16 @@ unsafe fn change_status_request_from_script_hook(boma: &mut BattleObjectModuleAc
                 }
             }
         }
+        // Checks whether you have landed on the first frame of non-tumble knockback
```
WuBoytH: why not have it check in the DAMAGE_AIR pre status script if the previous status was crouching
There are almost certainly scenarios where you CC a move and get lifted up for more than a frame?

jobrien97: > why not have it check in the DAMAGE_AIR pre status script if the previous status was crouching There are almost certainly scenarios where you CC a move and get lifted up for more than a frame?

u right, just pushed an update

WuBoytH (approved):

yoonikoo:
The nontumble hitstun landing adjustments make CC a real boy now. This makes it easier to punish nontumble options and approaches. CCing weak moves that don't put opponents into tumble do deserve to have counterplay against them. This also makes dragdown confirms on the ground weaker (Falco for example),

PR #1074, opened 2022-10-02
Perfect Pivot adjustment

jobrien97:
- Late Perfect Pivots now apply the same speed as normal Perfect Pivots (0.75 * dash speed), rather than less (0.5 * dash speed)
- Makes them even more consistent across different controllers


yoonikoo:
Tested on both GCC and handheld. It is indeed more consistent across different controllers and it makes sense that a late perfect pivot is the same speed as a normal one. It should not be punished for doing an already hard input. LGTM and is ready to merge.

WuBoytH:
Closed by request.

PR #1076, opened 2022-10-03
Dr. Mario - slight adjustments

jobrien97:
- Fixed an issue where Dr. Tornado could gain height multiple times in one airtime
- Removed Super Sheet heavy variant (Electric Blanket)
- Adjusted hitlag on electric normals
- Dtilt outer hitbox hit attribute changed: coin -> electric


yoonikoo:
Overall, really healthy changes. No reason for Dr. Mario to have had that infinite down special bug. In addition, removing the heavy variant of cape tones down Dr. Mario's oppressiveness. Down tilt feels so much better than the coin hitbox and replacing it with the electric variant, allowing for more versatility. I don't know that changed with the hitlag on his electric normals. I would need a change log in order for me to compare the two (nightly vs. this PR) side-by-side.

WuBoytH (approved):

PR #1086, opened 2022-10-06
Fix Chrom Double Edge Dance aerial momentum

black-calculus:
Updates Chrom's aerial side special momentum behavior to function like Roy's, which it should have been since the update to Roy's side special
Params chanegd:
 - Double Edge Dance initial aerial vertical speed decreased 1.2 -> 1.0
 - Double Edge Dance horizontal momentum multiplier increased 0.715 -> 0.8

### DIFF
--- a/fighters/chrom/src/opff.rs
+++ b/fighters/chrom/src/opff.rs
@@ -139,6 +139,18 @@ unsafe fn soaring_slash(fighter: &mut L2CFighterCommon) {
     }
 }
 
+pub unsafe fn double_edge_dance_vertical_momentum(fighter: &mut L2CFighterCommon){
+    let fighter_gravity = KineticModule::get_energy(fighter.module_accessor, *FIGHTER_KINETIC_ENERGY_ID_GRAVITY) as *mut FighterKineticEnergyGravity;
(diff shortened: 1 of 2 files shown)


yoonikoo:
The Double edge Dance aerial momentum fixes definitely make a difference in Chrom's Gameplay. Having it similar to Roy's makes its utility even better. SH Side B 1 can link into NAIR 1 or FAIR. Combo routes have become more versatile. Also, wavebouncing makes this move have more momentum than before.

jobrien97:
Can't this be put into his special_s init status in status.rs?

black-calculus:
Back when I initially rewrote the script for Marth and Roy I attempted that, no dice. Had to resort to putting the FighterKineticEnergyGravity shenanigans in opff, haven't been able to fully diagnose the issue yet

WuBoytH:
If you were calling original in the init status, you'd have to make sure the code ran after the original call.
If it *was* after the original call and the momentum stuff still didn't work, idk.

jobrien97:
But like...it's already entirely rewritten, there's no use of the `original` macro. And there's already a use of `SPECIAL_STALL_USED` within there.

WuBoytH:
Clearly I did not look

hdr-balance-backroom:
APPROVED

jobrien97:
Can we please check if this is doable in status scripts again?

techyCoder81:
> Can we please check if this is doable in status scripts again?

IMO this could (and perhaps even *should*) be done at a later time if its important to you. However, we do know that this implementation is fine, since its identical to the other FE chars.

techyCoder81 (approved):

PR #1090, opened 2022-10-08
Pichu Changes pt. 3

### LINKED ISSUE #1082: Pichu charge not granted by certain moves and bug in training mode
Of Pichu’s moves that are intended to give charge equivalent to the percent they deal, FSmash, Bair, and Dair have sort of issue with them. 
FSmash and Bair only award charge for the first hit of their multi hit, which does 2% (no staling). 
Landing Dair awards no charge, despite doing 4% (no staling).

When training mode is reset, Pichu’s charge amount is not reset. Additionally, if Pichu has charge and training mode is reset, Pichu keeps charge without a visual indicator, as if it had never been removed. Leaving training mode resets charge.

H3ra-57:
Even more goodies for our favorite little rat
Changelog: [Pichu_changes.txt](https://github.com/HDR-Development/HewDraw-Remix/files/9792729/Pichu_changes.txt)
resolves #1082 

For testers: please take note of how much Charge State activates in your matches, I want to know whether or not it activates too often now


yoonikoo:
Pichu is now a menace. Charge state definitely feels a lot better as an install option as more things now are interactable with the state. However, electroweb may be a bit too strong as stuns for a pretty long time. In addition, getting into charge state, in my opinion, is a bit too fast. Playing against Kaiokendo in ditto's really proved that. Zippy Zap feels rewarding, especially because you are now able to confirm into a charged up aerial for a kill option.

Kaiokendo661:
These changes are indeed promising in that they will make pichu even more of a unique character. However, I feel that pichu gets charge a bit too fast. Its very easy to get it at least 3-4 times in a stock unless you have very low openings-per-stock. Also, as Yooni said, Electroweb has an insane amount of stun for it to be balanced. Everything else seems perfectly fine imo!

yoonikoo:
LGTM with the adjustments and the electroweb effect changes!

WuBoytH (approved):

hdr-balance-backroom:
APPROVED

PR #1092, opened 2022-10-09
Enables AB Smash to work with tilt attack button

blu-dev:
woops i named the branch wrong lol.

anyways this enables tilt attack buttons to work like attack buttons for AB Smash (which is actually Attack + Special smash)


yoonikoo:
This will make having AB smash attack viable as a controller scheme. Tested it on DH and it is possible to still confirm into frisbee shot and gunman activation out of an AB smash attack.

WuBoytH (approved):

PR #1094, opened 2022-10-10
Remove extra hold frames on smash attacks

blu-dev:
Discussed with balance team ahead of time.

This is a weird change that Ultimate did in general, not to mention making it vary per character per move (with the most common being for 2 whole seconds past the last animation of charge).

However, since it is character specific if we want to restore it for Corrin, specifically, we could (maybe not keep it 2 seconds though...).


blu-dev:
We had a discussion about this in the Discord server, that begins here: https://discord.com/channels/659964948365049887/950977465000534086/1029093474932379678

TL;DR: It's important to remove this because sitting on ledge with a 10 million year threat duration because of holding your smash attack is ass.

O1yyy:
meeeeeeeeerge

WuBoytH (approved):

PR #1095, opened 2022-10-10
Add .gitattributes file to standardize autocrlf

jobrien97:
Makes .xml/.prcxml diffing not a nightmare

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Sets autocrlf for all contributors
+* text=auto


techyCoder81 (approved):
for clarity, this isnt an xml/prcxml-specific issue. This kind of wack diff would have occurred anytime someone changed a text file of any kind without having autocrlf set on their local installation.

blu-dev:
Needs more comments imo

PR #1096, opened 2022-10-10
Adding deletion lists for nightly-beta switching

techyCoder81:
Updating nightly and beta github CI to upload deletion lists for nightly-beta switching in the launcher. The result is that using verify as a hack to delete unnecessary files when switching between nightly and beta will no longer be an issue. Instead we will delete specific files we know should be deleted. This means verify will not need to be run when switching, which will make the switch significantly faster on both platforms.


WuBoytH (approved):
Please get these github actions out of my face

PR #1101, opened 2022-10-11
Allow Taunt Footstools to High-Jump

WuBoytH:
Before, taunt footstools would force a small jump. Now you can hold Taunt to jump higher.


yoonikoo:
Big sicko change. Now to super Mario jump on peoples' heads and footstool them to death off stage! LGTM

jobrien97 (approved):
lgtm

PR #1103, opened 2022-10-11
Increase jostle on dash grabs

### LINKED ISSUE #1000: Add some jostle on dash grabs
Less than vanilla, but more than none

jobrien97:
- Increases jostle overlap rate for dash grabs from 0.3 -> 2.0, to prevent whiffing at reasonable ranges

Resolves #1000


yoonikoo:
It feels so good to actually grab people without whiffing right in front of them. LGTM.

WuBoytH (approved):

PR #1107, opened 2022-10-13
Disable Aerial Jumps while Floating

WuBoytH:
Floats used to accept up inputs on the stick and burn your double jump in the process. This was especially prevalent with Dark Samus and Mewtwo, who could float upwards.

This fixes that so double jumps are simply disabled during all types of floats. To double jump, you have to stop floating first.

https://user-images.githubusercontent.com/26860994/195723387-884e13ad-ab88-454c-b839-8ae34df2fd56.mp4

(I don't have the visualizer in this clip but I promise I'm mashing the left stick up for every jump except for the instant floats)


stingers135:
Tested with Dark Samus, able to hold Up during float with tap jump enabled and retain double jump until after float ends. Seems to be working as intended.

black-calculus (approved):
lgtm 👍

PR #1108, opened 2022-10-14
Held Jab Removal

MadVillainLordQ:
Removal of held jab from every character in the cast other than Villager, Isabelle, and Pikachu along with implementation of jabs that have not been installed. No other hitbox properties were changed.


WuBoytH (dismissed):
That was probably hella tedious
looks good though

WuBoytH (approved):

hdr-balance-backroom:
APPROVED

PR #1110, opened 2022-10-14
change flick y impl for dash drop

blu-dev:
fixes the flick y impl for plat drop during dash, run, run brake, and turn dash.

before, if you moved your stick *anywhere* beneath -0.66 while you are performing one of those actions on a platform, you will drop through the platform. This makes crouch out of run on a platform borderline impossible.

this changes it to use an actual flick within the last 2 frames


yoonikoo:
Another QoL change. This makes dashdancing into dropping down from platform much more efficient than before. LGTM.

HisashiASM (approved):
ballz

PR #1111, opened 2022-10-14
Piranha Plant - Poison Leaves, hitbox fixes

Dune172:
Nair, Jab, Up Special, and Down tilt now have a short poison effect. Adjusted damage accordingly.

Fixed Down Tilt hitbox

Reduced dash attack sweetspot kb slightly.


H3ra-57:
My two cents, as one of the people working on the rework:

- Dtilt's outermost hitbox could probably just be removed (with ID 2's size reverted)
- You may also want to consider a nerf on uair to make it combo worse (bkb increase, kbg decrease so it doesn't kill too early)
- Part of the rework plan is to make Plant into a stance fighter. One of these is a poison stance, where Plant will deal poison damage on its bites, rather than its leaves. I would say axe the poison effect on the leaves

Otherwise, seems fine to me looking at it, looks like a good bandaid patch for things that are a bit overtuned

PR #1137, opened 2022-10-17
Marth redesign [TENTATIVE, NEEDS TESTING]

black-calculus:
After much consideration of Marth's current somewhat problematic state in HDR, I've decided that the best way to take this direction of Marth is to more closely replicate Melee/PM's tuning of his stats and key normals.

While not the only issue, a big factor of the relative lack of difficulty in Marth's current punish game and neutral lies in both how his damage and knockback is currently tuned. His damage earlier on was tuned down from Melee/PM levels due to how strong his combo game was in order to compensate for the powerful punish game he possessed. However, an unforseen circumstance of that was that while knockback was compensated given the relative reduction in damage from Melee/PM, the fact that each individual hit leaves you at comparatively lower percent than it would in Melee/PM coupled with his excellent frame data meant that his combos actually ended up being longer and more frustrating to get hit by as you would not travel quite as far after each hit, allowing him to set up more powerful and favorable resets mid combo and trap you in an even stronger punish game than was intended and than he can achieve in Melee/PM. This is coupled with the higher BKB to attempt to curtail early percent combos, which actually ended up backfiring as he's much safer on hit and has more frame advantage at early percents, meaning his exceptionally strong punish game could start even earlier.

On top of this, HDR Marth has had faster air speed, run acceleration, air acceleration, and maximum horizontal jump speed than Melee/PM, which, albeit only a slight bit higher, has served to make his punish game and neutral even stronger.

Overall, piling all of these attributes on top of the already strong design that Marth has in Melee/PM makes for a quite problematic character that has been an issue in the public eye since HDR's inception, and whose problematic factors I haven't quite been able to identify until my recent examination and testing of his attributes and tools.

Keeping all of these in mind combined with our recent introduction of ASDI and strengthened crouch canceling, I have decided to experiment with retuning Marth's stats and key normals to be overall much more similar to Melee/PM, while retaining some of the key attributes from HDR that make Marth unique here such as Dancing Blade and jab.

**Utilt:**
The increase in damage and KBG and the reduction in BKB as well as the readjusted angle should serve to make the move require a bit more effort to follow up on, while still retaining combo utility Marth must use his movement much more effectively to capitalize on the DI you choose after getting hit by this move. The damage and KBG increase both serve to make the move a more rewarding anti-air as well as provide it a bit of killing utility at high percents while making it a bit less of a confirm tool at later percents to encourage him to choose other options to combo with.

**Dtilt:**
Not much to say here other than the reduction in BKB on the inner body hitboxes both make them set up more scramble situations as well as makes CC a bit more viable against this move. Given how they have lower priority than the middle blade hitbox it won't come up that often, but I figured I'd match the tuning/design of this move to how it is in Melee/PM since it's already 75% of the way there.

**Nair:**
The reduction of KBG on the second hit serves to make the move more appealing combo filler at all percents and makes the move feel a bit more natural to use, allowing him to use it in slightly more creative ways. This change, while also having other implications with the rest of his kit, serves to make this move more appealing to use over fair in more instances mid-punish and in neutral, especially in tandem with the changes made to fair in order to make his neutral and punish game a bit more engaging.
The normalization of the tipper hitboxes with the rest of the blade is quite tentative and more of an experimental change, which I found to feel a bit more natural in terms of utilizing the move. I am considering returning the stronger hitbox on the tip to reward good spacing with the move to be able to net you kills with the tipper hitbox, but more testing will be needed on that front.

**Fair:**
The reduction in BKB to fair and increase in damage should serve to make the move much more bearable to fight at all percents, allowing you to ASDI/CC it at low percents to combat careless usage of the move, and make the move less of a default combo filler at mid to late percents given you'll fly farther after each individual fair which will provoke Marth to use his movement more intelligently in tandem with landing the move in order to continue his punish game.

**Bair:**
The reduction in BKB on the sourspots is mainly to provide Marth better access to the reverse bair conversions he can utilize in Melee/PM to make the move a bit more fun to mess around with mid-punish.

The changes I'm working on should make Marth much more engaging to both play as and against, as he is in Melee/PM. While the issue of insane frame data is still there, it's a big part of what makes Marth's neutral and combo game fun so retuning his moves' reward on hit is a much healthier solution.

Full changelog here:
[Marth_Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/9803426/Marth_Changelog.txt)


Marleyan:
So far the changes are pretty good and definitely more in line with what Marth should be. I know it's currently a wip and there's more changes coming, but I wanted to give some thoughts/feedback and show some findings around up tilt that I believe could make it a bit problematic. 


https://streamable.com/h97mov
https://streamable.com/1rslig
https://streamable.com/l5g63b
https://streamable.com/sn10rc
https://streamable.com/p3945m
https://streamable.com/5a2p1e
https://streamable.com/s55b4t

with these showings in mind (I can provide more if necessary) it seems that the changes to up tilt give it increased KBG that could be a bit problematic as it results in quite ridiculous early kills on no di with no real setup, and also seems to kill combo potential almost entirely at mid-high% rather than simply just encouraging marth to utlize his movement more to net rewards from it. 

800px-Utilt_with_text although up tilt in the pr is much more in-line with this, it's still not quite accurate as the hitboxes for up tilt in melee send at different angles depending on whether it's the early or late hitboxes of the move. (Which not only encourages utilizing movement to get punishes, but also requires marth to react more depending on if he hits the late\early part of the move, rather than simply just arm/innersour/sour/tipper structure)  Also, I'd suggest maybe lowering the KBG or reverting both the KBG and BKG to how they were previously, as the numbers in the current wip pr are 1:1 to melee but it's well known that melee numbers do not always accurately transfer well to HDR and Marth's up tilt probably needs further experimentation to accomplish what you're going for.

SkewedAskew:
Resolved in https://github.com/HDR-Development/HewDraw-Remix/pull/1980

PR #1147, opened 2022-10-21
Aegis changes

black-calculus:
Some cleanups to Mythra's stats and normals and a few readjustments to Pyra's normals to move towards making the pair flow together as a cohesive set together.

Changelog here:

[Aegis_Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/9883472/Aegis_Changelog.txt)

[Will post hitbox comparisons once made]


stingers135:
Putting my full thoughts on all the changes here after playing for a couple of hours from a balance perspective:

**Pyra**
Jab: Didn't really notice the changes, seems fine

Fsmash: Didn't really notice the changes, seems fine

Fair: AC fair is great, its not overcentralizing because there are plenty of situations (honestly probably more often than ac fair) that I want to land with fair for more pressure on shield. AC fair fills a unique niche in her gameplan and I think its a slam dunk.

The KBG increase is also noticeable, I felt myself getting side kills from much closer to center stage than before. I approve but keep an eye on this if Pyra becomes too strong as an obvious balance lever. 

Bair: Still seems plenty strong to me, changes seem fine

Uair: Didn't notice myself killing much earlier with it despite the large increase in KBG for whatever reason. But I would keep an eye on this one, might be overcentralizing in floaty matchups (didnt play vs. any floaties in my playtests)

Dair: Could probably use more sourspot nerfs, still too strong IMO

Upb: I like the hitbox changes, the last one felt a little silly even though it wasn't really OP

**Mythra**
Attribute changes: She feels buttery smooth. Great work

Jab: Didn't notice the differences, seems fine

Ftilt: Small difference but it was noticeable, doesn't change too much in my exp but no complaints

Utilt: Feels much better, the changes here make it followup better and feel better

Dtilt: Didn't notice the differences, seems fine

Dash attack: This one is a slight miss IMO, I think that the higher angle is useful in certain scenarios but sending out was also useful in other scenarios. This one makes it weaker overall because the opponent has more control over DIing a combo setup move than a move designed to send them offstage...in a perfect situation this new DA is stronger but when you're fighting a competent opponent that DIs properly that advantage turns into a disadvantage over the old one. I suggest an angle difference on the sourspot (tipper), outer hitbox is generally more useful sending outwards since if you're reaching that far to hit them your odds of following up are lower on average anyway.  That should get a revert to sending offstage, sweetspot (hilt/mid) hitbox can stay as current one since it can be very useful for followups to catch bad DI or low/mid% opponents. Also, the fact that you're committing yourself deeper into them with the close hitbox means that you should have a higher reward for actually landing the move, ie. potential for followups. Seems like a counterplay win and a win for those using her

Fsmash: Angle change seems pretty minimal in impact unless you catch someone slipping but that seems like a good spot for this move to be. I like this change overall

Dsmash: Not sure how I feel yet, it seems like it'd be good for countering CC if CC were actually useful but its too weak in this game to matter. DACDS is stronger than before in general...I think this in combination with the dash attack change seems superfluous? Like they both kinda do the same thing now. I think this can stay if the dash attack changes I suggested go through though since its unique. But if you don't want to change DA then I don't see a world where this stays in its current state...in that scenario I'd add the horizontal gimp hitboxes back to tipper hit on dsmash so she still has the old usage of it available with good spacing.

Dthrow: I think the Angle change kind of cancels the impact of the KB changes, seems like on good DI its still hard to follow up. Still don't see any reason to use this over Uthrow (generic combo throw) or fthrow (at ledge to get offstage)

Nair: Considering this in tandem with the changes to other aerials this is a slam dunk. It fills a niche as a combo ender now.

Fair: This one is weird, against mid% opponents that DI in you actually cant follow up because they don't go far enough for you to reach them when factoring in your aerial momentum. Which is unique counterplay...I suppose it could stay but I'd prefer if that weren't an option? I'd prefer if the counterplay was DIing out like normal because then they're putting themselves at risk by putting themselves further offstage...in its current stage it seems like a bad move against competent opponents. Would suggest higher BKB

Bair: Slam dunk

Uair: Don't really notice the changes, seems fine

Dair: Slam dunk

Upb: Are you sure that "initial slash additional hitstun decreased" change went through? If I'm reading it right seems like it's suggesting that you get less hitstun on upb? Because I don't really notice a difference. One side effect of the 6f change on blade switch is that you can mythra upb > instant switch and not leave the ground, which I actually think is really cool. I'm pretty sure it confirms into Pyra usmash on no DI which is kind of OP but you can't confirm on DI out (or up I think) so I think it's okay to stay...being able to stay grounded after it actually adds a shitton of additional options for us, it's also fairly technical of a technique to pull off since you need to blade switch immediately after confirming the upb (didn't to testing to confirm the exact frame # but its a fairly tight link), I think it's a fun and unique tech that adds a lot to her game so I hope it stays. But in general the upb changes in and of themselves are fair, the hitbox being shrunk is definitely noticeable, but it was obscenely strong before and still remains a cornerstone of her gameplan so it definitely deserved the nerf in that regard. 

Also this is a personal thing that I don't think would be a big deal one way or another, but could ray of punishment have infinite range? The word infinite sounds OP but in all honesty I don't think it would make a huge difference, in 90% of cases when your upb wouldve hit them but doesn't only because it timed out, they were probably dead anyway. I just reaaaaaaaaaally hate getting a sick angle on the upb hit and it missing because the range is just a little too short, lol

Blade switch: Alluded to, but incredible that you were able to shorten the length of that. Very fun change, great work

**Bugs**
Didn't notice any bugs that were brought on by these changes (I think) but I have fallen through the map numerous times while playing on this PR. I think this is an Aegis bug that was introduced with the ECB changes, but in case it was anything you did here try to doublecheck anything that might've impacted that. But it seems like Mythra or Pyra in tumble are very susceptible to falling through the map currently, during playtests there was one game where it happened three times in just that one game.

hdr-balance-backroom:
APPROVED

WuBoytH (dismissed):

black-calculus:
Changelog updated with some final adjustments for now taking into account initial feedback.

PR #1157, opened 2022-10-25
ECB overhaul

### LINKED ISSUE #1146: ECB issues strike again!
Clipping through stages still happens, only encountered with Rainbow Cruise. 

https://fxtwitter.com/CrossHariko/status/1582968992616296448?s=20&t=UyNqz8FBzY3ghsIcozcoWA

### LINKED ISSUE #1158: Sephiroth Going Through Stage
If you blade dash edge cancel and immediately go into b reverse octoslash, Sephiroth will clip through the stage if he holds a direction into the stage (2 videos attached). Uncharged will slightly clip through the stage but will always result in Sephiroth landing on stage (video of this is shown with me doing 75 degree angle).

https://user-images.githubusercontent.com/116753613/198100820-5616234f-42ae-4384-929e-9a5fb23aa756.mp4

https://user-images.githubusercontent.com/116753613/198101091-75fb8733-4efc-4654-aee4-900961f7f11c.mp4

https://user-images.githubusercontent.com/116753613/198101201-398c21d1-af17-4ca6-be17-f6caa1a8e5b3.mp4

jobrien97:
**ECB shifts now behave identically to Melee:**
- For the first 9 frames of being airborne, your ECB's bottom point will be locked to your Top bone (your base position).
- On frame 10 and onwards, your ECB's bottom point will use its regular ECB pattern, meaning it will be calculated identically to the other 3 points.

Previously, we had used a hard "ECB shift amount" constant, defined per character, to shift your ECB upwards.
Now, we allow the game to calculate the bottom ECB point identically to the other 3 points, which means its location can vary significantly depending on how an animation contorts your character's body in the air.

Marth's double jump (old):

https://user-images.githubusercontent.com/47401664/197868717-f85ab207-c21c-4527-aa69-3b6108591641.mp4

Marth's double jump (new):

https://user-images.githubusercontent.com/47401664/197868741-e63a5018-5720-4ec1-9b93-c593b91c2449.mp4

 ---
The previous implementation shifted your *entire* ECB upwards, rather than just the bottom ECB point. This caused your character's head collision to be raised as well, which may have caused some weird ceiling interactions.
Now, your ECB is "squished" upwards from the bottom point, to around your character's knees, with the top point remaining tied to your head.

Frame 12 of Marth's jump (old):

ecb_old

Frame 12 of Marth's jump (new):

ecb_new

 ---
The new implementation also allows for the return of **No-Impact Lands (NILs)**:
- Due to how the game handles collision, it perceives "moving downwards" as any time your current frame's ECB bottom is lower than the previous frame's ECB bottom. The game only allows you to land when "moving downwards."
- A character's aerial animation may contort their body in such a way that their ECB's bottom stretches downward for a frame or two. During this time, the game perceives you as "moving downwards," **even if you are moving upwards**. This means the game allows you to land.
- Thus, if you are just above a surface when your ECB stretches downwards, *even if moving upwards*, you will land with your character's normal landing lag (2-5 frames depending on the character).

Here is one (frame-perfect) example in action:

https://user-images.githubusercontent.com/47401664/197866360-5d2d0835-f456-4a36-9259-87ef525514f7.mp4

This also fixes platform canceling aerial attacks.

Resolves #1146 
Fixes #1158


yoonikoo:
Platform cancelling indeed is back. Tested it on Marth, Falco, Joker. There seems to be no ECB issues relating to Ridley or Yoshi anymore with this PR. LGTM.

WuBoytH (dismissed):

dynamic/src/consts.rs
```diff
@@ -226,6 +226,7 @@ pub mod vars {
             pub const ATTACK_DASH_ENABLE_AIR_LANDING: i32 = 0x1005;
 
             pub const SHOULD_WAVELAND: i32 = 0x1000;
+            pub const DISABLE_ECB_SHIFT: i32 = 0x1111;
```
blu-dev: no

fighters/common/src/function_hooks/collision.rs
```diff
@@ -98,36 +98,89 @@ unsafe fn get_touch_flag_hook(boma: &mut BattleObjectModuleAccessor) -> i32 {
...
+// This function calculates the coordinates of the passed bone relative to the Top bone (PostureModule::pos)
+// It stores these x/y/z coordinates in param_3's Vector3f
+
+// ground_module_ecb_point_calc_hook will pass each bone from your character's map_coll_data list in vl.prc, one by one, into this func
+// If param_6 in ground_module_ecb_point_calc_hook is 1, it will then pass the Trans bone once all of the map_coll_data bones have been processed
+// The game will use your Trans bone's distance from your Top bone to determine where to place your bottom ECB point, which will pretty much always be {0, 0, 0}
+// This is why your ECB bottom point is always "locked" to your Top bone
+
+// By returning once the Trans bone is passed into this func, we can ignore it and thus use your map_coll_data bones to calculate your bottom ECB point, like the other 3 points
+#[skyline::hook(offset = 0x48fc40)]
```
blu-dev, line 169: I don't like this, for a number of reasons:

Any point *ever* we call this function with `trans`, it's going to return invalid data and that's not really going to be documented anywhere.

That being said, I would appreciate if this was handled via a VarModule flag to wrapped around the `call_original!` in the above hook. I don't want to be chasing a bug 10 months from now related to this. It's the simplest solution, something like:
```rs
VarModule::on_flag((*boma).object(), vars::instance::IS_GETTING_POSITION_FOR_ECB);
call_original!(...);
VarModule::off_flag((*boma).object(), vars::instance::IS_GETTING_POSITION_FOR_ECB);
```
and then check the flag in this hook.

blu-dev (changes requested):

techyCoder81 (approved):
The changes made seem to meet the requests in previous review. lgtm.

PR #1162, opened 2022-10-30
Sora changes

Wish325:
- Jump Distance 1.665 -> 1.495
Sora was clearing Captain falcon distance with his dash jump, which is too far considering the Keyblade.
+ Upthrow angle changes 
    88 -> 108 
+ Downthrow angle changes
    78 -> 74
    The purpose of these is to make these throws mix-ups into each other, correct di for down throw is incorrect for up throw and Vice Versa
+ Thunder summons on frame 14 from 18 (hitbox now appears on frame 18)
Testing a change to improve it's overall utility.
 
- Down Smash frame 20 -> 15

- Jab 2 Has been adjusted to prevent downtilt confirms at low percents


hdr-balance-backroom:
APPROVED

techyCoder81 (dismissed):

techyCoder81 (approved):

PR #1166, opened 2022-10-31
Return the ability to grab ledge with a half circle motion during runoff

techyCoder81:
This un-removes the ability to grab ledge from stage using a quick half-circle motion during ledge runoff 
(to be discussed internally)


blu-dev:
This needs in depth discussion by the balance team first, do not merge this without a comment from @hdr-balance-backroom

hdr-balance-backroom (changes requested):
After much deliberation, we have decided that this should not be implemented. There have been many developments in character-specific ledge grabbing tech, and having this "one-size-fits-all" method of easily and optimally grabbing ledge would detract from their viability

PR #1167, opened 2022-10-31
Buff wavedashes universally to halfway between before and now

techyCoder81:
This buffs wavedash lengths (`landing_speed_mul_escape_air_slide`) universally to halfway between what they are now, and what they were before.
Originally: `0.957`
Currently: `0.832`
This PR: `0.8945`

From here on out we intend to balance this on an individual level.


blu-dev:
Can you change this to have a global value in `common.xml` that we get via `ParamModule`, and then change all of these values to 1?

techyCoder81 (commented):
I am approving this since blujay ended up making the changes instead

PR #1182, opened 2022-11-03
Wavedash/waveland landing speed readjustment

black-calculus:
Readjusts how landing speeds during wavedash/wavelands work:
Our airdodge speed magnitude (2.79) currently matches Melee/PM's value, but given that you're unable to wavedash/waveland fully horizontally in those games unlike in HDR, that meant that our maximum length wavedashes with full landing speed went much too far until we implemented reduced speed on airdodge landing. Currently, your landing speed is multiplied by the global wavedash landing speed multiplier `wavedash_speed_mul`, (which atm is 0.8945). This was implemented to keep wavedash and waveland speeds in check for our maximum length wavedashes/wavelands. However, this affects _all_ angles of airdodges, including diagonal angles as well as full horizontal, making diagonal angles a good bit less useful than in Melee and PM. This was remedied a bit by the recent PR increasing that landing speed multiplier, but I personally felt that angles below max could be improved slightly

This PR makes it so that you retain full landing speed from an airdodge, but the absolute maximum horizontal speed you can achieve from a wavedash/waveland is capped at `escape_air_slide_speed * wavedash_speed_mul` (which with the current values would be `2.79 * 0.8945`), which is the current maximum wavedash/waveland speed you can achieve right now (if your wavedash/waveland angle would make you go faster than that, your speed is capped off at that value). This keeps maximum distance wavedash lengths in check, but makes angles below that more suitable for medium/moderate microspacing while being less stiff.

The results will vary per character, but as a quick example for reference, here's the change in Marth's diagonal notch wavedash (tested on a Gamecube Controller):

Before:
Diagonal (Pre-Testing)

After:
Diagonal (Post-Testing)


fighters/common/src/general_statuses/airdodge.rs
```diff
@@ -96,10 +96,11 @@ pub unsafe fn status_end_EscapeAir(fighter: &mut L2CFighterCommon) -> L2CValue {
             WorkModule::set_float(fighter.module_accessor, landing_frame, *FIGHTER_INSTANCE_WORK_ID_FLOAT_LANDING_FRAME);
             let global_speed_mul = ParamModule::get_float(fighter.object(), ParamType::Common, "wavedash_speed_mul");
             let speed_mul = WorkModule::get_param_float(fighter.module_accessor, hash40("param_motion"), hash40("landing_speed_mul_escape_air_slide"));
-            let speed_mul = speed_mul * global_speed_mul;
+            //let speed_mul = speed_mul * global_speed_mul;
             fighter.clear_lua_stack();
             lua_args!(fighter, FIGHTER_KINETIC_ENERGY_ID_STOP);
-            let speed_x = app::sv_kinetic_energy::get_speed_x(fighter.lua_state_agent) * speed_mul;
+            let escape_air_slide_speed_clamp = WorkModule::get_param_float(fighter.module_accessor, hash40("escape_air_slide_speed"), 0) * global_speed_mul;
+            let speed_x = (app::sv_kinetic_energy::get_speed_x(fighter.lua_state_agent) * speed_mul).clamp(-escape_air_slide_speed_clamp, escape_air_slide_speed_clamp);
```
WuBoytH, line 103: The `clear_lua_stack` and `lua_args` lines should be directly behind the line that needs it (`sv_kinetic_energy::get_speed_x`) for readability purposes.
I would even take `sv_kinetic_energy::get_speed_x` out of the equation and store it in a local variable first, though that might just be a me thing

WuBoytH (changes requested):

WuBoytH (approved):

PR #1184, opened 2022-11-03
revert b reverse window to 10f

techyCoder81:
this reverts the b reverse window from 15f back to 10f, to reduce unintended B reverse inputs.


blu-dev (approved):
lgtm

PR #1185, opened 2022-11-04
Villager changes!

SuddyN:
[Villager_Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/10072661/Villager_Changelog.txt)

# CHANGELOG 🟢🟡🔴

### Sapling 🟢
🟢 Can now be B-Reversed in the air
🟡 Sapling plant hitbox f7->f13
🟢 When using **DTilt** & **DSmash** near sapling:
* The sapling is removed
* The move's hitbox is replaced with a hitbox on the sapling:
  * DMG ->15 (DTilt) 20 (DSmash)
  * Angle -> 90
  * KBG -> 92 (DTilt) 82 (DSmash)
  * BKB ->68 (DTilt) 55 (DSmash)
  * Unique sounds and effects are added

https://user-images.githubusercontent.com/33226440/203668703-40cb312a-3084-4469-bb2a-86a0434851a5.mp4

### Watering Can 🟢
🟢 Can now be B-Reversed in the air
🟢 Can now be canceled with the shield input
🔴 Hitbox removed (not the windboxes)

https://user-images.githubusercontent.com/33226440/203668741-54cd4a56-1c19-4e99-9529-b07a6780f45c.mp4

### DSmash 🔴
🔴 Can no longer have an optional bury hitbox

Devs have previously expressed distaste for moves with multiple optional functions. The DSmash optional bury was already niche, and being buried sucks as a player experience.

### DTilt 🟡
🔴 Overall hitbox size reduced
🟢 Main hitbox extended to cover the grass

image
image

### FTilt 🔴
🔴 Sweetspot KBG 100->82, BKB 60->31, Angle 361->22
🔴 Sourspot KBG 40->82, BKB 81->31, Angle 40->25
🔴 Furthest hitbox brought in to match the animation

image
image

Old FTilt was notorious for killing at ridiculously low percents, as its sweetspot had a ton of knockback. I've brought that knockback down a lot, but to make sure the move still sees some use, I've reduced the angle and increased base knockback. 
Now, FTilt is better at sending into tech situations and forcing players to ledge.

### Up Special 🟢
🟢 Can now be canceled with Attack or Shield to cancel into free fall (not special fall)
🟡 Costs a majority of fuel to do so. Actionable 6f after starting cancel, and can't start cancel until 23f after starting UpB (so as not to allow cancel combos)

### Grab 🟢
🟢 Standing grab f14->f10
🟢 Dash grab f16->f13
🟢 Pivot grab f16 -> f14

### Dash Attack 🟢
🟢 Slightly larger early hitbox now covers villager's arms up to his shoulders 

Before
After

### BAir 🔴
🟡 Final Hit Sweetspot/Sourspot added. (Both hitboxes kill later than pre-patch, but sweetspot is much closer to pre-patch)
🔴 Sweetspot DMG 8.5->8.0
🔴 Sourspot DMG 8.5->7.0
🔴 Sourspot size reduced
🟡 Hits one more time
🟢 Multihit damage 0.5->1.5 per hit
🔴 Dragdown hitstun reduced (KBG 80->48, BKB 50->35)

BAir is like a swiss knife; a good combo tool, fast startup, fast recovery, multihit, dragdown tool, and a good kill move. In order to emphasize other parts of villager's kit and reward good spacing, overall knockback has been reduced, and a sweetspot has been introduced.

Before
After

### NAir 🔴
🔴 KBG 105->95
🔴 Hitbox size reduced
🔴 Late hitbox starts 1f sooner

image
image
image
image
image
image


SuddyN:
OTG Grab will be coming later, after I make sure the bugs are ironed out :)

If this PR gets reviewed before then, it'll be in a separate PR

SkewedAskew:
Very thorough for a first contribution! Initial thoughts:
- Nair could use a hitbox cleanup. Try a single hitbox on Villager's center for the first two frames, then once he splays his arms add a hitbox on each hand and resize/reposition the existing hitbox to his waist area to cover the legs
- Ftilt's outermost hitbox does not line up with the animation, changing the Y-offset of id2 from 9.0 to 8.0 should fix that
- Dtilt's main hitbox does not extend to where the grass is being pulled from the ground, feel free to extend the hitbox ~1.5-2.5 units to cover that area if you want. In addition, consider moving both hitboxes up a couple units as they are positioned very low
- For the new sapling dtilt:
	- It should be possible to make the strong hitbox spawn at the sapling's position instead of on Villager's body, which would make more sense for this variation
	- Instead of an explosion, try a sys_crown effect
	- Just an idea, perhaps you may want to extend this functionality to down smash as well in some way?
- A long-standing issue is that back air has very high frame advantage on hit when used as a dragdown (as high as +16). To correct this, you'll need to decrease the BKB and KBG of the multihits. Somewhere around the 18-28 range for both values is probably what you're looking for, but you'll need to feel it out
- If you're feeling particularly adventurous, Villager could use the shield cancel on up special that Isabelle has (not the bloondash part)

SuddyN:
- [x] Update the changelog
- [x] Nair adjustments
- [x] Ftilt nerf
- [x] Dtilt's main hitbox extension
- [x] Bair dragdown nerf

For the new sapling dtilt:

- [x] Better hitbox positioning
- [x] sys_crown effect
- [x] Extend this functionality to down smash
    
Additional features:

- [x] Watering can canceling

Won't do:

- Instead of UpB canceling I implemented watering can canceling. I don't think it would be smart to give villager shield cancel up B. Watering can canceling differentiates villy from isabelle, but it has the potential to be powerful. I think villy should suffer the weakness of having a bad up B, or gain a new/unique up B like the rising lloyd rocket suggested by @HisashiASM 
- If you think canceling up B is OK, I would suggest giving him the ability to cancel it directly into airdodge but not regular shield cancel

SkewedAskew:
Round 2:
- Sapling dtilt/dsmash doing 20% and killing at ~80% is a bit much, especially considering its potency around ledge. It may be fine with just decreasing damage to around 16% and hitlag mul to 1.5. Additionally, it needs more oomph effect-wise; try adding something like `sys_attack_impact` and extending the `sys_crown` to the entire horizontal range of the sapling hitbox
- Nair is still very disjointed for a move of its ilk, and being attached to neck bone may introduce certain Z-axis inconsistencies. Try attaching it to top bone and centering it using y and z offsets. The initial hit can be slightly beyond Villager's body, but the late hit should not travel past his limbs
- Id2 on the final hit of bair should be about 1.5-2.0u smaller, but the damage can be increased a little (it only deals 10% total)
- Your checks in opff for down special cancel can be simplified by using `boma.is_status_one_of(&[*STATUS_1, STATUS_2, etc])`

MadVillainLordQ:
After discussing, we have the following notes:
- Grab improved from frame 14 to frame 10 (if you do implement being able to grab knocked down opponents then we'll have to discuss)
- Up b cancel mechanic similar to Isabelle with a reduction in fuel
- We prefer Ultimate's Nair style hitboxes to the singular hitbox you currently have. I'd recommend going with the same hand and hip hitbox sizes that ultimate has along with the additional head hitbox in HDR with a matching hitbox size to the hip
- With shield cancellable moves, we generally want them to have some sort of commitment. In the case of water can, it would be within our design philosphy to make it shield cancellable starting on frame 5 (the frame after water pops out) and having some minimal lag showing the can is being put away. About 4 frames should do it. I'd recommend looking at how Yink's dair was coded so that you can fast forward the animation to putting away can when holding shield.
- Although sapling dtilt seems fairly difficult to land, we'd prefer if at least now a distinction can be made between dsmash sapling and dtilt sapling in which dtilt sapling damage is reduced to 15%. At the very least, it would make sense given the commitment, that dsmash sapling is stronger than dtilt sapling.
- We'd like to see some sort of visual effect that better demonstrates sapling removal. I know explosion was removed but perhaps you can try using the pocket's visual effect instead.

SuddyN:
- [x] Grab improved from frame 14 to frame 10
- [x] Up B cancel mechanic
- [x] Ultimate-style NAir hitboxes
- [x] Can canceling adjustment
- [x] Sapling DTilt adjustment
- [x] Sapling effect adjustment
- [x] Changelog updated

hdr-balance-backroom (dismissed):
APPROVED

hdr-balance-backroom:
APPROVED

SkewedAskew (approved):
Lgtm

PR #1190, opened 2022-11-04
Fix landing momentum when landing on slopes

### LINKED ISSUE #767: Falling on slopes is slippery / will transfer your landing momentum
To reproduce:
Roll back to the edge of a stage with slopes like Yoshi's Story and short hop in place. You will slip off ledge due to the fact that you slide in the direction of a slope when landing

jobrien97:
Your character will no longer slide in the direction of a slope when landing on it. They will land as if they had landed on a flat surface, as it functions in PM.

Resolves #767


SkewedAskew (dismissed):
Slopen't

PR #1191, opened 2022-11-04
Dash/Turn/Run Turn tech reimplementation

### LINKED ISSUE #1104: Clean up moonwalk logic
- Move to control energy funcs
- Implement P+ mechanic: disable dashback if stick falls below certain threshold

jobrien97:
Tech accessible during Dash/Turn/Run Turn has been moved into the game's internal energy functions. This includes:
- Moonwalks
- Pivots/Perfect Pivots
- Sticky Walking
- Boost Running
- Bidou/Aidou/C-Stick moonwalks
- Shield stops

This vastly cleans up the implementation of these techs, and allows them to be a more "natural" byproduct of our engine, rather than being tacked-on.

While this alone may not feel any different in-game than the previous implementation, there have been several QoL changes made to movement alongside this:
- Implemented the PM feature which disables dashbacks if your stick falls below a certain threshold (-0.5) during dash
   - This prevents certain methods of moonwalking from triggering accidental dashbacks
- Lowered the pivot stick threshold from -0.8 -> -0.6
   - This should make pivots more consistent to perform across different controllers
- Late Perfect Pivots now travel the same distance as frame-perfect Perfect Pivots, rather than using a lower speed multiplier

Resolves #1104


blu-dev (dismissed):
approved, should be tested and probably merged into nightly

SkewedAskew (approved):
Lgtm

PR #1195, opened 2022-11-05
Pichu - Discharge buff

jobrien97:
Minimum power/size of Discharge is now 50% of full strength rather than 25%.
Effect sizes for reduced power Discharge also adjusted to better match hitboxes.

Also moves Pichu's Charge State constants to ParamModule.


SkewedAskew (approved):
Lgtm

hdr-balance-backroom:
APPROVED

PR #1201, opened 2022-11-11
Jigglypuff - on-hit Rest visual change

jobrien97:
As most people know, Jigglypuff has an earlier FAF when she hits a Rest, compared to a whiffed Rest. There was previously no indicator as to when this on-hit FAF was reached, meaning Jigglypuff could act deceptively out of her sleeping animation.

Now, Jigglypuff skips to her wake-up animation before her on-hit FAF is reached, to allow for proper timing for punishes.


WuBoytH:
The previous commit btw allows for Rest to have its earlier FAF if Puff successfully lands a hit, but also hits a shield at the same time.

WuBoytH (approved):

hdr-balance-backroom:
APPROVED

PR #1202, opened 2022-11-11
Projectile behavior fix

jobrien97:
Certain projectiles should no longer stick to the ground without despawning if reflected/deflected at an angle.


techyCoder81 (approved):
this seems like it should work. Hard to get tested.

PR #1207, opened 2022-11-16
Robin Nair/Thoron tweaks

MadVillainLordQ:
Minor correction on Robin's Nair startup and Thoron buff
- Previously Robin's nair was supposed to matched vanilla ultimate's frame data but did not. Robin's nair now currently starts at frame 7 instead of the previous frame 9.
- Thoron's startup has been reduced from frame 23 to frame 17. The grounded/aerial FAF has been reduced from 75/78 -> 61/63


hdr-balance-backroom:
APPROVED

SkewedAskew (approved):
Lgtm

PR #1211, opened 2022-11-19
Fixes the meter UI and related control menu crash

techyCoder81:
- fixes meter ui showing up on the wrong player in certain circumstances
- fixes the control menu crash, which was due to meter ui


SkewedAskew:
Verified the controls menu no longer crashes

SkewedAskew (approved):
Lgtm

PR #1213, opened 2022-11-21
Hitfall Mode, Airdash Mode, and multi-mode selection for custom game modes

techyCoder81:
- adds a hitfall gamemode next to turbo and tag.
- adds an airdash mode.
- adds the ability to toggle multiple modes at once


techyCoder81:
it would be cool to also have an airdash mode

utils/src/game_modes/airdash.rs
```diff
@@ -0,0 +1,73 @@
...
+use smash::phx::Vector3f;
+use crate::util;
+
+pub unsafe fn update() {
+    // skip this frame because the match hasnt started
+    if !app::sv_information::is_ready_go() {
+        return;
+    }
+
+    println!("doing airdash update!");
```
WuBoytH: We probably shouldn't have these printing every frame if we can help it

WuBoytH (changes requested):

utils/src/game_modes/airdash.rs
techyCoder81: great catch, thanks

WuBoytH (approved):

H3ra-57:
Works on my end 👍

PR #1214, opened 2022-11-21
Byleth Adjustments

MadVillainLordQ:
Byleth has received some damage and angle adjustments along with a significant rework to up special. Further refinements are needed to be done with up special, but that will be done in the future.

Notes:
- Byleth can only grab opponents with up special once per airtime (including if this is done from grounded up special). If Byleth up specials in the air after the first usage, the up special will have a weak launching hit similar to how Min-Min's up special functions.
- Byleth's tether height when connecting with grounded up special is much closer to the ground than prior (about two training blocks difference)

[Byleth Old vs. New Hitboxes](https://imgur.com/a/2Dm6CUg)
[Byleth PR Changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/10061259/Byleth.PR.Changelog.txt)


O1yyy:
the changelog txt is one of chrom’s lol, not byleth

hdr-balance-backroom:
APPROVED

jobrien97:
UpB can still spike, its behavior seems to have been reverted to vanilla, it spikes after a certain percent threshold regardless of input

SkewedAskew (approved):
Lgtm

PR #1221, opened 2022-11-28
Allow users to set control schemes without the system stopping them

techyCoder81:
Currently, if you attempt to set a control scheme without using the Attack option somewhere, the game will stop you from saving. You cannot set Tilt Attack on a button and Smash Attack on cstick, which should be a perfectly valid scheme, without the game blocking you. This should fix that behavior.

Needs testing.


techyCoder81:
does not perform as expected, required button assignments cannot be controlled via our existing lua

PR #1230, opened 2022-11-30
Collision improvements

jobrien97:
Removes a leniency with collision that Ultimate had in place, which, when coupled with the new ECB changes, allowed you to land on platforms while rising (platform cancel) pretty frequently. Platform cancels should be less frequent now, and more intentional.

Also fixes any remaining visual bugs when landing (characters appearing stuck halfway into ground on first frame of landing).

### DIFF
--- a/fighters/common/src/function_hooks/change_motion.rs
+++ b/fighters/common/src/function_hooks/change_motion.rs
@@ -14,9 +14,16 @@ unsafe fn change_motion_hook(boma: &mut BattleObjectModuleAccessor, motion_hash:
     let mut start_frame = arg3;
     change_motion_pos_shift_check(boma);
     if boma.is_fighter() {
-        // Starts heavy landing animation on frame 2
+        // Starts landing animations on frame 2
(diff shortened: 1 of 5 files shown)


techyCoder81 (dismissed):
looks about as reasonable as I could expect

techyCoder81 (dismissed):
seems about as reasonable as I could expect

techyCoder81 (approved):
seems about as reasonable as I could expect

yoonikoo:
LGTM. Tested with Calc and saw no issues.

yoonikoo:
Ridley has an issue at the apex of nair on Hollow Bastion.

yoonikoo:
https://user-images.githubusercontent.com/105788245/205430928-03ea3dfb-383f-439d-86a3-96c6ef32f78b.mp4

techyCoder81:
for better or for worse, this is not surprising, as the new behavior *will* still allow for animation-shifted NILs, they should just be less common. Merging this behavior based on testing and approval.

PR #1243, opened 2022-12-05
Villager Followup Adjustments

MadVillainLordQ:
Further adjustments have been made to a recent Villager PR. This includes a rebalancing of balloon trip that now has the ability to be cancelled into an actionable state, removal of turnip rng, and a few hitbox property changes.

Notes:
- Turnips are no longer random for uair/dair and are now on set cycle going from 1-2-3
- Villager can no longer wall jump
- Balloon trip height on max fuel has decreased significantly, while its height on empty fuel has been marginally increased
- Sapling now requires six drops of water to become a tree, up from 5 drops. Interval between water drops from sprinkler has been decreased from 6 frames to 5 frames. This was done as a temporary fix to prevent sapling from growing in 2 instant sprinkler cancels. Currently, after immediately cancelling a sprinkler, using sprinkler again will form a drop on frame 1 of the move allowing for 3 drops to appear in what normally should only allow 2 drops. This helps mitigate against that by requiring an extra water drop if that method is used. Conversely, holding sprinkler will now do 6 drops so functionality of the held sprinkler has not changed. 

[Old vs. New Hitbox Comparisons](https://imgur.com/a/SgG7kgx)
[Villager PR changelog.txt](https://github.com/HDR-Development/HewDraw-Remix/files/10270944/Villager.PR.changelog.txt)


SkewedAskew (dismissed):
Lgtm

hdr-balance-backroom:
APPROVED

SuddyN:
BAir effects and sound effects play before the hitbox. Can cause some weirdness on trades and early dragdown attempts. 

https://user-images.githubusercontent.com/33226440/208787858-5390a442-922a-454f-96ba-6d432f5ef2c5.mp4

SuddyN:
The watering can watering speed adjustment and tree growing requirement changes don't prevent the use of two instant can cancels from growing a tree, other than the first time. The first two cancels will give 2, 3 drops. Then, the next two give 3, 3 drops which is enough to water the tree in two cancels. The watering can is reset whenever can is used without canceling it.


https://user-images.githubusercontent.com/33226440/208793986-3c51f289-2203-4b53-999c-2cf6249086ca.mp4

SkewedAskew (approved):
Lgtm

PR #1254, opened 2022-12-13
Remove electric move knockback smoke farts

jobrien97:
Knockback smoke does not generate until electric hitlag is over


WuBoytH (changes requested):
Looks reasonable, I'd prefer you comment out the code and say "Now Unused" or something like that instead of just flat out removing it
Do you have a video of the before and after?

WuBoytH (approved):

PR #1255, opened 2022-12-14
Mii Swordfighter - remove heavy tornado variant

jobrien97:
Though it was supposed to be removed already, heavy tornado is still accessible currently, while being graphically identical to regular tornado. This removes it properly.


WuBoytH (approved):

hdr-balance-backroom:
APPROVED

PR #1261, opened 2022-12-19
Angle-able Sonic Blade

WuBoytH:
Allows Sora to angle Sonic Blade 30 degrees up or down on the first rep.  Also removed the homing properties from Sonic Blade.

https://user-images.githubusercontent.com/26860994/208331486-a4284a96-9d19-45f5-b593-db52a3521090.mp4

Notes:
- The solution for making the first rep angle-able was basically telling the game we were already on the second Sonic Blade. Because of this, Sora uses `special_s2`, `special_s3`, and `special_s3` again, instead of the usual `special_s1`, `special_s2`, and `special_s3`.
- This means that `special_s1` is now unused, though I've left it in the code. This also means that, despite increasing the `attack_num` param in his vl.prcxml from 3 to 4, he can still only perform 3 Sonic Blades. I would fix this, but a full reimplementation of that status is the worst case scenario.
- In addition, the `attack_speed_x` param was increased from 3.2 to 3.5 due to each subsequent Sonic Blade becoming slower. This is to make it feel like the original Sonic Blade.


MyFriendAaron:
I tested this change and it feels awesome. Great work WuBoy! A truly Herculean task.

hdr-balance-backroom:
APPROVED

SkewedAskew (approved):
Lgtm

PR #1266, opened 2022-12-25
Properly Aggregate Nana BattleObject When Getting All Players

SuddyN:
Problem
=======

In methods which need the battle objects of all player fighters, a for loop through entry ids 0..8 is used. In the case of Ice Climbers, they need to get Nana separately. Each function does this to varying levels of success, resulting in odd behavior with Nana.

Solution
========

Implements a new utility function which gets the battle objects of all active player fighters, including a special case for Nana. Replaces existing `for i in 0..8` loops with loops over the ids aggregated by this function.

- Fixes Nana getting ledge trumped every time
- Fixes Nana not using turbo in turbo mode
- Needs testing in all custom game modes, and weird ledge situations


SuddyN:
Just realized there is redundant code in ledge and change_status files, where they individually try to account for Nana. That code can be removed, and I'll get to it tomorrow.

dynamic/src/util.rs
```diff
@@ -154,6 +154,30 @@ pub fn get_active_battle_object_id_from_entry_id(entry_id: u32) -> Option<u32> {
     }
 }
 
+/// includes nana
+pub unsafe fn get_all_player_battle_object_ids() -> Vec<u32> {
```
techyCoder81, line 158: does this also work for aegis, since they are technically different battle objects too?

SuddyN: If I'm correct, it gets the active battle object, which is whichever of the two characters is currently out. Works for use cases of this function because you wouldn't need both. This should be tested.

techyCoder81 (approved):

PR #1277, opened 2023-01-03
Allow three intangible ledge grabs and remove ledge grab limit

techyCoder81:
Allows refreshed iframes on ledge grab, up to 3 ledge grabs. Subsequent grabs will not have iframes, and the hard limit of six ledge grabs per airtime has been removed (it is now 999).


SkewedAskew (approved):
Lgtm

PR #1278, opened 2023-01-03
Banjo Overhaul Fixes

CSharpM7:
-Unintentional movement from electric attacks removed. Fail flag is now only set during SideSpecial start status
-Small code cleanup: includes moving state-specific functions outside of OPFF , and consolidating Beakbomb OPFF functions


MadVillainLordQ (approved):
LGTM

PR #1280, opened 2023-01-03
Removes the grab timer

techyCoder81:
Reduces the grab timer from 35f -> 1f


SuddyN (approved):
tested and works - one line change, lgtm

PR #1282, opened 2023-01-04
Spikes cause knockdown splat rather than ground bounce

techyCoder81:
With this change, spikes will cause a knockdown splat, rather than a ground bounce.


kalomaze:
are we absolutely 100% doing this or is it still under contention

techyCoder81:
we held an internal vote on it and it was pretty heavily in favor of trying this, at least for a while

MadVillainLordQ (approved):
LGTM

PR #1283, opened 2023-01-05
simplifying backroom approval process

techyCoder81:
simplifying the backroom approval process and in the process reducing the amount of traffic across the github API, hopefully resulting in reduced api limit issues.


WuBoytH:
I... assume this is the kind of thing where you'd need to get rid of the workflow on the main branch to function as intended?
From what I understand the PR approval comment shouldn't be there.

PR #1286, opened 2023-01-06
Bayonetta Jab123 Fix

Detronlegacy:
Bayonetta steps forward considerably during her Jab123. With the current angle, the opponent is not pushed forward with her, which can have her step past them even with no SDI. This simple fix corrects that issue.

Jab1: Changed angle of id's 0-2 from 75 to 50
Jab2: Changed angle of id's 0-2 from 75 to 50
Jab3: Changed angle of id's 0-2 from 65 to 40


techyCoder81 (approved):
very simple acmd changes

PR #1287, opened 2023-01-07
Fix Byleth up special command grab still spiking

black-calculus:
Fixes an issue where Byleth's up special would still spike when used after activating the special input forward smash (the `IS_HEAVY_ATTACK` flag wasn't properly being cleared at the beginning of the up special footstool ACMD script, so when turned on via activating special input forward smash that flag would persist into the up special status if regular forward smash wasn't used beforehand)


SkewedAskew (approved):
Lgtm

PR #1288, opened 2023-01-08
System Changes (+ Wolf adjustments)

blu-dev:
**SYSTEM CHANGES:**
- Sped up main menu load times
- Removed music reset on training mode reset

**WOLF CHANGES:**

Side Special:
- Removes spiking sweetspot
- Sourspot becomes only hitbox, size nerf from 7.0 -> 5.5 (previously the sourspot was 7 units and the sweet spot was 4 units)
- Changed effect to remove the side swipe claws
- Decreased active frames from 5 -> 2
- Removed blindspot and added better trailing hitboxes
- The actionability out of side special only occurs when hitting the launcher hitbox, on either shield or hurtbox
- Removed the inconsistent input window for illusion shorten
- If wolf is in the "rush" part of the status, pressing B will cause shorten to occur and the launcher hitbox to spawn the next frame. This is not bufferable, there are only two frames you can get a shorten. One is the early shorten and the other is the late shorten, and if you miss both of those the move finishes to completion

Before/After (Shown: Active frame 2 so interpolation is accounted for)
Wolf Special S BeforeAfter

Down Tilt:
- Body hitbox:
    - Size: 5.0 -> 4.0
- Leg hitbox:
    - Size: 4.0 -> 3.5
- Foot hitbox:
    - Size: 5.0 -> 3.5
    - Angle: 75 -> 80

Before/after:
Wolf Dtilt BeforeAfter

Neutral Special:
- Removed bayonet hitboxes
- Moved blaster bullet spawn x position from 7 -> 0 in order to catch the deadzone between tip of gun and wolf

Bugfix
- Removed ability to b-reverse landing


SuddyN (commented):
FYI this makes the sweetspot have worse knockback than sourspot according to https://www.desmos.com/calculator/p1wybtguqp

jobrien97 (approved):
lgtm

PR #1291, opened 2023-01-08
Ground bounce adjustments

jobrien97:
- Removes the 0.8 hitstun multiplier applied to ground bounces
  - This 0.8 multiplier still applies to ceiling/wall bounces (stage spikes)
- Increases knockback speed threshold required to trigger ground bounce 0.8 -> 1.0
  - As a result, splats will occur more often
- Reduces ground bounce knockback speed multiplier 0.9 -> 0.7
  - As a result, characters won't bounce as high off the ground


blu-dev (approved):
lgtm

PR #1293, opened 2023-01-09
Mewtwo Hurtbox Fix

Ultimegiddo:
8a7f2730e634b53ea35e7e2425c71015
Before
2c62101ca7f8f54aebaa8bedd5625249
After


HisashiASM:
Sorry but no, we don't see this being fit as this would cause a disjoint too large for this title. HOWEVER I am going to leave this PR open.

Instead I want you to make it always on, EXCEPT for Dash and Dash Turnaround. This will not be in VL anymore, but instead ACMD Intan.

techyCoder81:
changes have now been made which enable the hurtboxes during all of his attacks which use the tail, such as ftilt, dtilt, uair, etc, so that they remain non-disjointed and can trade, while removing the annoying tendency of this character to sometimes get snagged in neutral states.

techyCoder81 (approved):

PR #1296, opened 2023-01-10
Add PeepeePoopoo back to DPit

HisashiASM:
Tpose was missing, so we fixed, push shit immediately.


hdr-balance-backroom:
Needs to feel more intentional and not buggy, change to a taunt input out of jump canceling down special :>

HisashiASM:
Imma be real here, the fact it happens is what makes it funny. Me, Job, and Blujay we’re all sad to see this go, I think this a a lore exception.

O1yyy:
rip bozo

PR #1298, opened 2023-01-13
Rosalina hitbox edits

FSFrozen:
Bair: Late hit terminates 5 frames sooner (ends on frame 16 from 21)

Uair: Motion rate added to end of animation (+4 endlag)

Utilt: Late hit terminates 2 frames earlier (ends on frame 15 from 17)


Z-hime:
Very good step in the right direction.
Luma kb thresholds and hp differences make it apparent that Rosalina now has to protect Luma much more than before. Matchups seem to feel a bit less one sided in the fact that Luma is able to be disposed of a bit easier now.

Uair changes are great to prevent several staircase combos against characters that may not be fast falling enough to deserve them. Overall felt much more enjoyable to fight as you can focus your energy more on Rosalina rather than trying to get through Luma walls.

LGTM!

fighters/rosetta/src/acmd/aerials.rs
```diff
@@ -126,6 +126,7 @@ unsafe fn rosetta_attack_air_hi_game(fighter: &mut L2CAgentBase) {
     }
     frame(lua_state, 45.0);
     if is_excute(fighter) {
+        FT_MOTION_RATE(fighter, 4.0);
```
hdr-balance-backroom: This motion rate should be set earlier in the move with a smaller multiplier to avoid strange animation inconsistencies

hdr-balance-backroom (dismissed):

hdr-balance-backroom:
Please include a full changelog in the main body once complete

SkewedAskew (dismissed):
Dash attack still feels too disjointed. As an idea, try changing the outermost hitbox to a set of two capsules like so:

image

Otherwise lgtm

SkewedAskew (approved):
Lgtm

PR #1301, opened 2023-01-14
Simon's Non-invasive Fix-Buff-Patch-Stuff

elias-michaias:
Todo:
- [ ] update this changelog

- Down Air's grounded spike KBG 35 -> 65
- Down Air's inactionable on-hit animation is now actionable after frame 15
- Neutral Special now takes 80% of the total time that it did to use
- Neutral Special can be fastfell during any part of the animation
- Up Special no longer locks your momentum, letting the player drift after using it

GOAL:
Fix current problems with Simon and make his kit function as intended with his current design philosophy. Nothing super crazy.


MosaMosa-x:
These changes make simon feel great and really give simon that spice he needed for a long time and address some long-time bugs he had without having to go overboard. Axe feels great to use now with its speed-up changes, although it's a bit challenging to use but synergizes well with his other projectiles enough to even it out.

I'm a big fan of the new dair, it gives him a lot more exciting stuff to land and will come in handy with footstools IMO. Up b having its momentum fixed is also huge now that you can actually use it on the ledge. Overall I'm a big fan and ran into no issues testing these out the last few days.

hdr-balance-backroom:
If you want to reinstate the ability to fast fall Cross, go for it

Riah321:
Utilt vacuums and both Utilt and Uair are too easy to convert off of (usually into themselves or each other) Utilt is also way too active, huge and safe on shield, hits low and to the side before and after the whip anim.  Dair has too little landing lag seeing as it's safe af on shield and can be easily edge cancelled. The spike on dair also seems unnecessary. Up throw up B was literally free way late into kill % range. Up special hit too far away from the fist, especially at the peak. This was part of what made up throw up air so easy With a lot of corner carry and easy reactions to DI.

He might benefit from being smaller, and by extension having smaller hitboxes on his body, making them more precise.

Idk who approved this but I'm sure they didn't play against a human.

techyCoder81:
sejong indicates that he does not intend to add back fastfall cross, therefore this PR is complete.

techyCoder81 (approved):

PR #1303, opened 2023-01-15
Bayonetta Fix Landing D-Air + KBG Nerf 

jaccob101:
Changes:
Down Air:
Landing active frames 0->2
Hitbox size 3.5 -> 4.0
KBG 127 -> 121

I noticed that when I daired on shield it only hit the shield once playing Bayonetta. When I checked frame by frame with the hitbox viewer I saw the hitbox was out 0 frames. My making it wait longer to clear hitboxes the move now properly comes out and links decently consistent after my change. I also nerfed the KBG because I feel that having the move work suddenly while it didn't before might have an impact on her balance. I do not know when this move was last changed or why the move did not work for so long.

-Still does not connect well if you hit them with a late hit dair but that sounds like a skill issue

https://user-images.githubusercontent.com/122749442/212571132-7db5a74b-2d36-4a89-91fc-46ecaa34a143.mp4

https://user-images.githubusercontent.com/122749442/212571255-06881b6c-64c8-4d56-b037-3a06ee7114a0.mp4


SkewedAskew:
Rolled into #1331

Issue #1302, opened 2023-01-14
Nerf dash attack distances


jobrien97:
Get ratio of `(HDR dash speed / vanilla run speed)` and apply to dash attack energy.

PR #1314, opened 2023-01-17
Megaman Adjustments

SuddyN:
# Mega Man Adjustments

Potential further changes, if MegaMan proves to still be too strong:
- Leaf shield adjustments
- Less hitstun on lemons

ToDo:

- [x] DTilt changes
  - [x] Remove the jump cancel
  - [x] Increase FAF
  - [x] Increase range
- [x] DAir damage decrease
- [x] Update the changelog

## Side Special
- can now fast fall during the move
- crash bomber manual detonation FAF 45->35

## DTilt
- jump cancel removed
- startup f4->f5
- early and late hitbox values combined into one hit
  - dmg 8
  - angle 77
  - kbg 70
  - bkb 70
  - effectively similar to the old early hit, but with less damage 

## UTilt
- removed the light version, now heavy is the only version
- kbg 103->75
- bkb 38->84
- late hitbox values reverted to vanilla
- final late hitbox begins f12->f9
- final late hitbox ends f20->f19

## Dash Attack
- travels 80% as far (accompanies #1302)
- startup f8->f10
- early to late hit transition happens 3f sooner
- faf 39->40
- early hit:
  - angle 40->361
  - hitlag 1.1->1.0
  - no longer reverse hits
- late hit:
  - angle 85->65
  - hitlag 1.1->1.0
  - no longer reverse hits

## FAir
- early/main hit:
  - kbg 100->89
  - bkb 35->53
  - tipper dmg 12->13
  - shoulder hitbox removed (also, it was previously overriding the hilt hitbox)
- late/lingering hit:
  - angle 361->70
  - kbg 110->64
  - bkb 12->35
  - tipper dmg 9->8
  - shoulder hitbox removed (also, it was previously overriding the hilt hitbox)
  - ends 2f sooner

After:

https://user-images.githubusercontent.com/33226440/214985913-0d4fe3d0-217d-4dad-9366-b256b623913e.mp4

# BAir
- multihit reverted to vanilla, but using 365 angle instead of 367
- launcher kbg 232->183

# UAir
- body hitbox removed
- new scoop hitbox on f9
- late autocancel f51->f40

New Scoop (1f before the projectile):
 
image

# DAir
- body hitbox removed
- f1,2 have new hitbox values:
  - dmg 16
  - angle 270
  - kbg 61
  - bkb 8

### DIFF
--- a/fighters/rockman/src/acmd/aerials.rs
+++ b/fighters/rockman/src/acmd/aerials.rs
@@ -33,7 +33,7 @@ unsafe fn rockman_attack_air_f_game(fighter: &mut L2CAgentBase) {
     if is_excute(fighter) {
         WorkModule::on_flag(boma, *FIGHTER_STATUS_ATTACK_AIR_FLAG_ENABLE_LANDING);
     }
-    frame(lua_state, 7.0);
+    frame(lua_state, 8.0);
     if is_excute(fighter) {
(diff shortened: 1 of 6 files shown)


hdr-balance-backroom:
Down tilt:
- Remove jump cancel. It's already a very fast sliding tilt (with leg intangibility)
- Increase endlag (currently 32)

Forward Air:
- Remove first active frame shown in video

Down Air:
- Reduce damage from 20 to 14-16

SuddyN:
I did already remove that first frame of MM FAir, sorry it's not in the notes yet :P
OK to restore (or partially restore) the travel distance of DTilt again, since the jump cancel is getting removed?

hdr-balance-backroom:
Sure, thanks for the update

PR #1315, opened 2023-01-17
Dash attack speed/distance adjustments

### LINKED ISSUE #1302: Nerf dash attack distances

jobrien97:
Adjusts burst speed and travel distances of dash attacks across the cast, based on the difference in movement speed between HDR and vanilla for a given character.

To do this, a "general movement speed" value was calculated for each character, which is the average of their initial dash speed and maximum run speed (`(dash_speed + run_speed_max) / 2`). This value was calculated for both HDR and vanilla, which were then used to produce a ratio, signifying the difference in a character's "general movement speed" between games.

This ratio was then applied as a speed multiplier to the character's dash attack, allowing it to be more adequately suited for their new HDR movement speed.

While this ratio was used as a base, some characters had further tinkering to their speed multiplier to make dash attack/DACUS/DACDS less overbearing.

This results in shorter dash attacks/DACUS/DACDS for most of the cast, while a select few are slightly longer.

Resolves #1302

### DIFF
--- a/fighters/bayonetta/src/acmd/ground.rs
+++ b/fighters/bayonetta/src/acmd/ground.rs
@@ -152,6 +152,7 @@ unsafe fn bayonetta_attack_100_end_game(fighter: &mut L2CAgentBase) {
 unsafe fn bayonetta_attack_dash_game(fighter: &mut L2CAgentBase) {
     let lua_state = fighter.lua_state_agent;
     let boma = fighter.boma();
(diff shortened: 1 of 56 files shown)


SkewedAskew (dismissed):
Lgtm

SkewedAskew (dismissed):
Lgtm

techyCoder81:
waiting on further adjustment/discussion of some proposed changes

techyCoder81 (approved):

PR #1318, opened 2023-01-17
Version info on arena

techyCoder81:
puts the currently installed HDR version information on the arena ui.

Arena Version Info


SkewedAskew (approved):
Lgtm

PR #1324, opened 2023-01-18
Alolan Whip cleanup

SkewedAskew:
Up:
~ [R] Changed startup windbox to match that of no angle
~ [R] Removed pull-in windbox; grab changed to air-only and made constantly active throughout normal duration

Down
~ [+] BKB: 85 -> 80

Also removes belt flare vfx on Cross Chop's flip


SuddyN (dismissed):
LGTM, and seems to achieve its intended purpose

techyCoder81 (approved):

PR #1335, opened 2023-01-24
fixes drift di for ground and air meteors

techyCoder81:
drift di is presently entirely nonfunctional during meteors/spikes, both grounded and aerial. This fixes that.


jobrien97 (approved):
lgtm, tested and works

PR #1336, opened 2023-01-25
Global air acceleration nerfs

### LINKED ISSUE #1329: Air acceleration adjustments
Some characters' air acceleration values are still close to vanilla, while their kits have been largely differentiated from vanilla.

jobrien97:
Brings air acceleration values closer in line to Melee/PM's average.

Also nerfs a select few characters' max air speeds:
- Dr. Mario 1.0465 -> 0.95
- Popo 0.875 -> 0.83
- Nana 0.91875 -> 0.87375
- Villager 1.09 -> 0.946
- Wii Fit Trainer 1.1 -> 0.977
- Rosalina 1.065 -> 1.0
- Palutena 1.02 -> 0.959
- Ridley 1.0675 -> 1.0
- King K. Rool 0.98 -> 0.906
- Isabelle 1.095 -> 0.978
- Terry 0.9775 -> 0.911
- Byleth 0.945 -> 0.853
- Min Min 0.9475 -> 0.85
- Sephiroth 0.95 -> 0.911

Resolves #1329


Riah321:
Who is being nerfed?
What is the rational for this set of changes?

O1yyy:
> Who is being nerfed? What is the rational for this set of changes?

image

Riah321:
I tested it with a few chars. Felt fine when going from ground to air, but felt rather artificially slow when descending after a fullhop, especially with characters that are fast on the ground. Some characters that depend on drifting back after a short hop aerial for shield pressure in neutral will suffer greatly from this patch. Perhaps just accel could be worsened and stable air speed could be left intact to preserve the naturality of how air momentum feels in HDR since the goal is to make drifting back and forth worse.

SkewedAskew (approved):
Lgtm

PR #1337, opened 2023-01-28
Zelda Din now properly despawns on input

HisashiASM:
Input side special while there is a dins out to the desummon the current projectile and summon a new one, WOOOOWOWOOWOOW.


SkewedAskew (approved):
Lgtm

techyCoder81:
labeling this as bug because it was the original intent that could not be fixed in the original PR

PR #1343, opened 2023-01-29
WaveLand and WaveDash Improvements

SuddyN:
Waveland and wavedash improvements via discussion between BluJay, Techy, and I - using Rivals as a point of reference, and adjusting for our engine and meta:

## TODO

- [ ] Reduce waveland/wavedash lag (currently 10f)
- [x] Allow "neutral" wavedash
  - If a wavedash is buffered while the stick is in neutral, we currently jump and then buffer airdodge. This allows a huge punish off a simple misinput. Instead, we could "neutral" wavedash in-place, which is functionally similar to a wavedash down from the ground.
  - Current behavior was originally implemented before we had Rivals-style airdodges, so its use case is no longer valid
- [x] Adjust the waveland snap threshold to always be from your characters' knees to their base position, even before the aerial ECB shift frame (first 10f in the air)
  - This should allow for more vertical leniency and consistency of wavelanding


SkewedAskew (approved):
Lgtm

yoonikoo:
LGTM. Feels great and smoovement.

PR #1346, opened 2023-01-30
Luigi Recovery Adustments

yoonikoo:
Bringing Luigi more in line with our future philosophy of recoveries.

Please replace the Luigi motion files from hdr-assets>fighters>luigi with the following folder: 
[luigi.zip](https://github.com/HDR-Development/HewDraw-Remix/files/10532253/luigi.zip)

<== LUIGI ==>
SPECIALS
-Down B (Cyclone)
~[+] Buoyancy_max increased from 1.3 to 1.45 (This allows Luigi to rise more with more mashing)
-Side B (Green Missile)
~[-] No longer grabs ledge
~[+] Brought back jumping side b momentum shenanigans :)
~[-] Stored misfire is lost upon death
~[-] Removed invincible frames on side b discharge

### DIFF
--- a/fighters/luigi/src/acmd/specials.rs
+++ b/fighters/luigi/src/acmd/specials.rs
@@ -426,7 +426,6 @@ unsafe fn game_specialairlw(fighter: &mut L2CAgentBase) {
     }
     wait(lua_state, 4.0);
     if is_excute(fighter) {
-        FT_MOTION_RATE(fighter, 0.8);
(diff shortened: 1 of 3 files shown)


jobrien97:
Looks like a motion rate was applied to downB's aerial endlag. Not sure if this was intentional, but if so, I don't think that's in scope for this PR

jobrien97 (approved):
lgtm

PR #1364, opened 2023-02-05
nerfs to some small things on dk

techyCoder81:
change notes:
- ground uspecial
  - inner hitbox size 8 -> 6
  - multihits damage reduced significantly
- air uspecial
  - extra shield damage removed (1 -> 0) per multihit
    - this move did a lot of shield damage, and this was why lmao
  - multihits damage reduced significantly
  - kb reduced kbg 85, bkb 60 -> kbg 80, bkb 40 (p+ values - damage is also the same as p+)
  - launcher hitboxes adjusted/simplified to a single capsule box
  - fixed autocancel bug (set yacl default on same frame as vanilla)
  - these changes mean that air (and ground) uspecial will deal far less damage
    and shield damage, and will be easier to directly challenge.
- fspecial armor duration reduced 14f -> 12f
  - this will reduce the ease of timing for  fspecial armor
  - this will also make fspecial more vulnerable to multihits. previously, there was only
    one frame between armor removal and hitbox creation, so sometimes multihits would still lose
    covering the space the previous 2 hitboxes filled
- pivot grab range reduced by 3 units
- if you attempt to airgrab and are holding an item, you will instead enter heavy item carry


blu-dev:
comments needed

techyCoder81:
> comments needed

huge, massive even

Riah321:
This PR fails to build. I was excited to test it too :(

SkewedAskew (approved):
Lgtm

PR #1366, opened 2023-02-06
Allow Neutral Wavedash

SuddyN:
Allow "neutral" wavedash (cherry-picked from #1343)

  - If a wavedash is buffered while the stick is in neutral, we currently jump and then buffer airdodge. This allows a huge punish off a simple misinput. Instead, we could "neutral" wavedash in-place, which is functionally similar to a wavedash down from the ground.
  - Current behavior was originally implemented before we had Rivals-style airdodges, so its use case is no longer valid


Riah321:
Seems fine to me after rather rigorous testing online.

PR #1369, opened 2023-02-07
Ground bounce threshold adjustment

jobrien97:
Knockback velocity required to trigger ground bounce (rather than a splat) increased 1.0 -> 2.5

Splats will now occur *much* more often (they most likely weren't occurring at all for some characters before), generally until 100-125%, but of course it will vary per move.

For reference, Captain Falcon's dair will cause Inkling to splat up until ~100%, after which Inkling will bounce.


yoonikoo:
LGTM. Best of both worlds.

Riah321:
LGTM :)
Olyyy agrees but he forgot to comment I think.

techyCoder81 (approved):

PR #1371, opened 2023-02-08
Waveland actionability changes

techyCoder81:
pulling changes from #1343 into its own PR


Riah321:
How many frames of landing lag have been shaved off? Knowing that is useful for testing and feedback.

Edit: It's gone from 10F to 5F.

PR #1372, opened 2023-02-08
Banjo hotfixes

jobrien97:
- Smoke gfx don't spawn if you roll off an edge
- Fixes Kazooie missing from ftilt lol


hdr-balance-backroom (approved):
Lgtm

PR #1385, opened 2023-02-15
Pichu - Discharge buffs

jobrien97:
- You now have 3 seconds after Charge State activates to use a full-strength Discharge before it starts decreasing in damage/size, rather than 1 second
- Discharge is now only 25% smaller at minimum size, rather than 50% smaller
- BKB increased 50 -> 70
- Your FAF now scales with Discharge's strength (lower FAF from weaker Discharge)

Minimum size before:
2023021314291100-0E7DF678130F4F0FA2C88AE72B47AFDF

Minimum size after:
2023021518182000-0E7DF678130F4F0FA2C88AE72B47AFDF


SkewedAskew (approved):
Lgtm

PR #1388, opened 2023-02-16
Random now includes DLC only if purchased

zrksyd:


blu-dev:
this ain't building bob

zrksyd:
What is the error? Does NNSDK have to be updated?

PR #1394, opened 2023-02-18
Remove magnifying glass shrinking

jobrien97:
In Ultimate, the closer you got to the blastzone, the smaller your magnifying glass would shrink. This was appropriate when paired with balloon knockback, but now that that is gone, this feature contributes to knockback feeling "slow" at times.

This reverts the behavior to the previous games': the magnifying glass's size now stays static.

Example:

https://user-images.githubusercontent.com/47401664/219884955-567d10e6-0325-4513-afff-821ff9a86cfc.mp4


black-calculus (approved):
lgtm chief 👍

PR #1397, opened 2023-02-19
Fast Fall on Neutral B

Nakedkoops:
What do you think, Mario has fast fall on neutral B (Fireball and Firebrand).


Nakedkoops:
Oh and btw I think fast fall fireball is a bit too good of a combo starter, just test it out to see if it's actually that broken.

PR #1405, opened 2023-02-23
Fix banjo dair bounce

jobrien97:
Fixes an issue where Banjo wouldn't transition to the bounce animation when hitting an opponent or shield


WuBoytH (commented):
Any reason is_infliction_status *wouldn't* be working here, or did OoO change something about how it works?

jobrien97:
> Any reason is_infliction_status _wouldn't_ be working here, or did OoO change something about how it works?

It does work, `set_frame_sync_animcmd` was just being called every remaining frame of dair's status after it hit, thus resetting to frame 45 of the anim every single frame

WuBoytH:
Ah I see

WuBoytH (approved):

PR #1411, opened 2023-02-24
Kirby - fix voice clip on landing hammer

jobrien97:
Fixes an issue where grounded sideB voice clip would play on uncharged aerial sideB landing.


WuBoytH (approved):
An odd solution but if it works...

jobrien97:
> An odd solution but if it works...

opff on main seems to have some interesting quirks ig

PR #1421, opened 2023-03-02
Rob changes

stingers135:
### Aerials
**Fair**: Reduced active frames by 7. Move now lingers for less time but is active earlier. Can do SHDF

**Bair**: Hitbox comes out on frame 20 and applies a short burst of movement (previously removed). Damage and kbg slightly reduced on strong and weak hitboxes. Centered innermost hitbox slightly.

Slowed down boost bair animation. Hitbox comes out on frame 28 and applies a larger burst of movement than before. BKB slightly reduced and KBG slightly increased (roughly normalized). 

Bair boost now has a charge system. Hold the attack button down to charge bair and make it move further and do more damage.

Reworked effect and sound timings to match new behavior. 

Bair boost is once per airtime

**Uair**: Reduced KBG and slightly increased BKB. Move will only kill if landed very high in the air, supposed to be used more as a combo/setup tool. In line with a more melee style philosophy. Does slightly more damage (1% more for each small hit)

### Specials

**Up B**: Added very slight burst of horizontal momentum when using. Largely unnoticeable just makes the move feel less stiff. Intangibility removed. Smaller height gain on aerial upb.

### Throws
**Uthrow**: Increased KBG. Should kill around 150~ on good DI against average weight/fall speed. Intended to compensate for Uair no longer functioning as a kill move. Seems healthier for the game in general.

### Tilts
**Ftilt**: Moved hand hitbox slightly further out, giving it a slightly longer disjoint. Should make move feel better. 

New: 
2023030220331600-0E7DF678130F4F0FA2C88AE72B47AFDF

Old: 
2023030220403300-0E7DF678130F4F0FA2C88AE72B47AFDF

**Utilt**: Decreased hitbox sizes slightly. Shouldn’t feel so free to scoop up anymore.

New: 
2023030220334100-0E7DF678130F4F0FA2C88AE72B47AFDF
2023030220335800-0E7DF678130F4F0FA2C88AE72B47AFDF
2023030220334600-0E7DF678130F4F0FA2C88AE72B47AFDF
2023030220340100-0E7DF678130F4F0FA2C88AE72B47AFDF

Old:

2023030220395900-0E7DF678130F4F0FA2C88AE72B47AFDF
2023030220400100-0E7DF678130F4F0FA2C88AE72B47AFDF
2023030220400400-0E7DF678130F4F0FA2C88AE72B47AFDF
2023030220400700-0E7DF678130F4F0FA2C88AE72B47AFDF


O1yyy:
great work for your first pr!!

fighters/robot/src/acmd/aerials.rs
```diff
@@ -222,65 +227,47 @@ unsafe fn robot_attack_air_b_game(fighter: &mut L2CAgentBase) {
                 WorkModule::set_float(boma, current_fuel - boost_fuel_depletion, *FIGHTER_ROBOT_INSTANCE_WORK_ID_FLOAT_BURNER_ENERGY_VALUE);
             }
         }
+		
+		if VarModule::is_flag(fighter.battle_object, vars::robot::status::BOOST_ATTACK){
+			FT_MOTION_RATE(fighter, 10.0/(8.0-3.0));
+		} else {
+			FT_MOTION_RATE(fighter, 0.7);
```
SuddyN: please use `FT_MOTION_RATE(fighter, num_frames / (end_frame - curr_frame))` for line 234, like on line 232

- end_frame is the frame you change motion rate next (or the last frame of the move)
- curr_frame is the frame you're writing this line on
- num_frames is the new number of frames you would like to get from curr_frame to end_frame

fighters/robot/src/acmd/specials.rs
```diff
@@ -1,58 +1,94 @@
...
+    FT_MOTION_RATE(fighter, 0.1);
+}
+
+#[acmd_script( agent = "robot", script = "game_specialairsstart", category = ACMD_GAME, low_priority )]
+unsafe fn robot_special_air_s_start_game(fighter: &mut L2CAgentBase) {
+    let lua_state = fighter.lua_state_agent;
+    let boma = fighter.boma();
+    frame(lua_state, 1.0);
+    FT_MOTION_RATE(fighter, 0.1);
+}
```
SuddyN: please use the formatting for FT_MOTION_RATE which I described in an another comment

fighters/robot/src/acmd/specials.rs
```diff
@@ -86,13 +122,30 @@ unsafe fn robot_special_air_lw_game(fighter: &mut L2CAgentBase) {
...
+		robot_special_air_s_start_game,
         robot_special_s_game,
+		robot_special_air_s_game,
         robot_special_s_end_game,
+		robot_special_air_s_end_game,
         robot_special_lw_game,
         robot_special_air_lw_game,
+        robot_special_hi_game,
     );
 }
```
SuddyN: consistent tab spacing pls

fighters/robot/src/opff.rs
```diff
@@ -18,6 +18,35 @@ unsafe fn gyro_dash_cancel(boma: &mut BattleObjectModuleAccessor, status_kind: i
...
+                    ControlModule::clear_command_one(boma, *FIGHTER_PAD_COMMAND_CATEGORY1, *FIGHTER_PAD_CMD_CAT1_AIR_ESCAPE);
+                    StatusModule::change_status_request_from_script(boma, *FIGHTER_STATUS_KIND_FALL, false);
+                    let vec = Vector3f{x: 0.0, y: -0.2, z: 0.0};
+                    KineticModule::add_speed(boma, &vec);
+                } else if boma.is_button_on(Buttons::Attack) {
+                    StatusModule::change_status_request_from_script(boma, *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_HI_ATTACK, false);
+                }
+            }
+        }
+	}
```
SuddyN: in my personal opinion, you should not be using `KineticModule::add_speed` to immediately allow ROB to grab ledge after a cancel, which is unlike any of the other characters with this type of recovery (who also can't immediately attack out of, or delay the climb of their recoveries)

however, below is a better re-writing of your method. Please avoid nesting `if` blocks, and instead use [guard clauses](https://deviq.com/design-patterns/guard-clause) or [logical-and](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND) to make conditional blocks easier to read and maintain.

```suggestion
    if WorkModule::is_flag(boma, *FIGHTER_ROBOT_STATUS_BURNER_FLAG_PUSH_B_BUTTON)
    && situation_kind == *SITUATION_KIND_AIR
    && frame > 11.0 {
        if boma.is_button_on(Buttons::Guard) {
            WorkModule::unable_transition_term_group(boma, *FIGHTER_STATUS_TRANSITION_GROUP_CHK_AIR_ESCAPE);
            ControlModule::clear_command_one(boma, *FIGHTER_PAD_COMMAND_CATEGORY1, *FIGHTER_PAD_CMD_CAT1_AIR_ESCAPE);
            boma.change_status_req(*FIGHTER_STATUS_KIND_FALL, false);
            let vec = Vector3f{x: 0.0, y: -0.2, z: 0.0};
            KineticModule::add_speed(boma, &vec);
        } else if boma.is_button_on(Buttons::Attack) {
            boma.change_status_req(*FIGHTER_ROBOT_STATUS_KIND_SPECIAL_HI_ATTACK, false);
        }
    }
```

fighters/robot/src/opff.rs
```diff
@@ -18,6 +18,35 @@ unsafe fn gyro_dash_cancel(boma: &mut BattleObjectModuleAccessor, status_kind: i
...
+// JC grounded sideb on hit
+unsafe fn jc_sideb(boma: &mut BattleObjectModuleAccessor, cat1: i32, status_kind: i32, situation_kind: i32, motion_kind: u64) {
+    if (status_kind == *FIGHTER_STATUS_KIND_SPECIAL_S || status_kind == *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_END || status_kind == *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_ATTACK) {
+        if (AttackModule::is_infliction_status(boma, *COLLISION_KIND_MASK_HIT) && !boma.is_in_hitlag()) {
+            if ControlModule::check_button_on(boma, *CONTROL_PAD_BUTTON_JUMP) {
+                boma.check_jump_cancel(false);
+            }
+        }   
+    }
+}
```
SuddyN: Please avoid nesting `if` blocks, and instead use [guard clauses](https://deviq.com/design-patterns/guard-clause) or [logical-and](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND) to make conditional blocks easier to read and maintain.

fighters/robot/src/opff.rs
```diff
@@ -163,7 +192,9 @@ pub unsafe fn moveset(fighter: &mut smash::lua2cpp::L2CFighterCommon, boma: &mut
     //gyro_dash_cancel(boma, status_kind, situation_kind, cat[0], frame);
 
     //neutral_special_cancels(boma, status_kind, situation_kind, cat[0]);
+    jc_sideb(boma, cat[0], status_kind, situation_kind, motion_kind);
     dspecial_cancels(boma, status_kind, situation_kind, cat[0]);
+	uspecial_cancels(boma, situation_kind, frame);
     bair_boost_detection(boma);
     fuel_indicator_effect(fighter, boma);
```
SuddyN: consistent tab spacing

romfs/source/fighter/robot/param/vl.prcxml
```diff
@@ -15,10 +15,12 @@
   </list>
   <list hash="param_special_s">
     <struct index="0">
-      <float hash="attack_speed_x">2</float>
-      <float hash="attack_speed_y">0</float>
-      <int hash="push_b_max">1</int>
-      <float hash="0x164fbe4f35">0</float>
```
SuddyN: instead of removing these params, lets set them to the vanilla values so they can be more easily adjusted in the future

SuddyN (changes requested):

stingers135:
Suggested code review changes added. One of those was a balance change so if the balance backroom goes over this let me know if that one is indeed required.

fighters/robot/src/opff.rs
```diff
@@ -18,6 +18,31 @@ unsafe fn gyro_dash_cancel(boma: &mut BattleObjectModuleAccessor, status_kind: i
...
+			WorkModule::unable_transition_term_group(boma, *FIGHTER_STATUS_TRANSITION_GROUP_CHK_AIR_ESCAPE);
+			ControlModule::clear_command_one(boma, *FIGHTER_PAD_COMMAND_CATEGORY1, *FIGHTER_PAD_CMD_CAT1_AIR_ESCAPE);
+			StatusModule::change_status_request_from_script(boma, *FIGHTER_STATUS_KIND_FALL, false);
+			//let vec = Vector3f{x: 0.0, y: -0.2, z: 0.0};
+			//KineticModule::add_speed(boma, &vec);
+		} else if boma.is_button_on(Buttons::Attack) {
+			StatusModule::change_status_request_from_script(boma, *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_HI_ATTACK, false);
+		}
+	}	
+}
```
SuddyN: this is purely styling, but there is some weird double-tab spacing here. Make sure your development environment is set up to convert tabs to spaces, because tabs are probably expanded differently on github than your environment

fighters/robot/src/opff.rs
```diff
@@ -18,6 +18,31 @@ unsafe fn gyro_dash_cancel(boma: &mut BattleObjectModuleAccessor, status_kind: i
...
+
+// JC grounded sideb on hit
+unsafe fn jc_sideb(boma: &mut BattleObjectModuleAccessor, cat1: i32, status_kind: i32, situation_kind: i32, motion_kind: u64) {
+    if (status_kind == *FIGHTER_STATUS_KIND_SPECIAL_S || status_kind == *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_END || status_kind == *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_ATTACK)
+    && (AttackModule::is_infliction_status(boma, *COLLISION_KIND_MASK_HIT) && !boma.is_in_hitlag())
+    && ControlModule::check_button_on(boma, *CONTROL_PAD_BUTTON_JUMP) {
+                boma.check_jump_cancel(false);
```
SuddyN, line 53: twice the needed spacing here too

fighters/robot/src/opff.rs
```diff
@@ -18,6 +18,31 @@ unsafe fn gyro_dash_cancel(boma: &mut BattleObjectModuleAccessor, status_kind: i
...
+			//KineticModule::add_speed(boma, &vec);
+		} else if boma.is_button_on(Buttons::Attack) {
+			StatusModule::change_status_request_from_script(boma, *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_HI_ATTACK, false);
+		}
+	}	
+}
+
+// JC grounded sideb on hit
+unsafe fn jc_sideb(boma: &mut BattleObjectModuleAccessor, cat1: i32, status_kind: i32, situation_kind: i32, motion_kind: u64) {
+    if (status_kind == *FIGHTER_STATUS_KIND_SPECIAL_S || status_kind == *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_END || status_kind == *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_ATTACK)
```
SuddyN:
```suggestion
    if ([*FIGHTER_STATUS_KIND_SPECIAL_S, *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_END, *FIGHTER_ROBOT_STATUS_KIND_SPECIAL_S_ATTACK].contains(&status_kind))
```

fighters/robot/src/opff.rs
```diff
@@ -160,12 +185,13 @@ unsafe fn fuel_indicator_effect(fighter: &mut smash::lua2cpp::L2CFighterCommon,
...
-    dspecial_cancels(boma, status_kind, situation_kind, cat[0]);
-    bair_boost_detection(boma);
-    fuel_indicator_effect(fighter, boma);
+	//gyro_dash_cancel(boma, status_kind, situation_kind, cat[0], frame);
+	//neutral_special_cancels(boma, status_kind, situation_kind, cat[0]);
+	jc_sideb(boma, cat[0], status_kind, situation_kind, motion_kind);
+	dspecial_cancels(boma, status_kind, situation_kind, cat[0]);
+	uspecial_cancels(boma, situation_kind, frame);
+	bair_boost_detection(boma);
+	fuel_indicator_effect(fighter, boma);
```
SuddyN: spacing

SuddyN (dismissed):

hdr-balance-backroom (changes requested):
- Increase fair endlag by 4 frames
- Revert all bair changes, try changing to a chargeable aerial similar to Corrin
- Remove nair and dair blue flame effects
  - those are used on bair as an indicator of fuel usage
- We've actively been moving away from having things be cancelable because it feels gimmicky, so these side b changes aren't going to happen. Instead, we'd like to see the move become a decently strong single hit
- Up b is interruptible too early, change the cancel frame from 11 to at least 18
- As compensation for the rest of the up b changes, up b needs its intangibility removed and dair will need its bounce to be restricted to once per airtime
- Lower uthrow knockback to kill around 150 with good DI
  - given how well ROB threatens shield with Gyro and potentially new side b, this throw shouldn't be nearly as strong as other kill throws

stingers135:
I'm fine with most of that but unclear on Bair. If I think of Corrin bair there's no "fuel cost" so are you thinking we just remove that whole implementation? I can picture this current bair being moved to the strong input and adding a new weaker bair that doesn't move you at all and just axing the whole fuel cost thing. It does feel a bit gimmicky but I thought if we were going to leave it it'd make more sense to go "all-in" on the idea.

And would you be fine with moving back shield cancel until 18 but leaving attack cancel in earlier? I think it helps a lot with making him more aggressive.

hdr-balance-backroom:
> I'm fine with most of that but unclear on Bair. If I think of Corrin bair there's no "fuel cost" so are you thinking we just remove that whole implementation? I can picture this current bair being moved to the strong input and adding a new weaker bair that doesn't move you at all and just axing the whole fuel cost thing. It does feel a bit gimmicky but I thought if we were going to leave it it'd make more sense to go "all-in" on the idea.

Similar to how Corrin's bair moves further the more it is charged, this new bair would also move further while consuming more fuel. Uncharged would be the same properties as the current light bair

stingers135:
I'll look at Corrins bair code, I thought it was just a tilt/smash input. It's important to me that ROB have access to a version of Bair that is both quick and changes his momentum. I don't particularly care about the fuel cost aspect of it.

 If your primary concern is gating it as a recovery tool by requiring fuel cost then my suggested implementation would be weak bair stays as I have it but uses a small amount fuel instead and gets moved to smash input, and then current strong bair would be the result of this charging effect.

SuddyN:
maybe the fuel cost could scale with the charge on the bair?

SuddyN:
> If your primary concern is gating it as a recovery tool by requiring fuel cost then my suggested implementation would be weak bair stays as I have it but uses a small amount fuel instead and gets moved to smash input, and then current strong bair would be the result of this charging effect.

sounds like way too much complexity on one move that way IMO. charge to gain power and mobility is more fluid and already has precedent

stingers135:
I just had an idea, if recovery was your main concern for bair, couldn't we do the same thing we're doing for dair and only give it one boost per airtime and then keep my current implementation?

stingers135:
I think I've hit a good sweetspot with these changes. Let me know what you think. Compared to first iteration:

Fair endlag increased by 4f

Side B reverted to nightly (probably won't end up changing this move at all as I don't have the capacity to realize my vision for it, the JC one wasn't really what I wanted to do anyway and I agree it was pretty stupid)

UPB attack cancel moved back 2f to 13. Upb shield cancel moved back to 20. Makes it higher commitment for ledge grabs but still viable for offense. 

To further nerf recovery, reduced vertical gain on initial aerial upb. This makes it a heavy commitment to quit using your upb if you are being pressured and have to cancel earlier than intended. I think this is a good middle ground between all of the options and lets him have both a very punishable but also very free form recovery. 

Intangibility and dair changes implemented as requested. 

Also changed bair to be one bounce per airtime but otherwise didn't change the move. I think this is a sweetspot between too useful for recovery that also enables it to be used in a more aggressive manner. My main concern with the charge idea is that to get useful horizontal momentum out of it, assuming uncharged is 0, heavily limits the ability to use it aggressively since you will spend too long holding charge to get real benefits from using it aggressively. Having a quick bair that can also give a small burst of movement lets you do things like this:

https://user-images.githubusercontent.com/115596132/224876318-8792d6ca-9037-4a15-a080-47270bd020cc.mp4

The inspiration behind this was Brawl ROB which was able to play in a similar fashion (though this is far nerfed from that iteration with a later upb cancel window, smaller hitboxes as well as about 7 less active frames on Bair making it less effective. Almost all of those bairs that missed there would've hit in Brawl) and I think the playstyle is much more interesting than how ROB currently functions with the more zoner styled archetype. Obviously you're never removing that completely since he has 2 strong projectiles but you can see how this enables alternate gameplans that are both interesting and nuanced. Please let me know if you find that agreeable since this is the main thing I was most passionate about with wanting to rework this character.

Uthrow nerfed as requested. This is the one I am least sure about because with how heavy the Uair nerfs are I am worried he will have trouble killing. We'll see

stingers135:
Added charge effect on Bair similar to Corrin's as requested. Modified effects to match. Modified fuel costs to change per level of bair used. Nerfed power of bair to compensate for increased control

stingers135:
Complete

hdr-balance-backroom (dismissed):
- Remove the added up special cancel, increase the ending gravity on ROB's normal cancel instead

hdr-balance-backroom (dismissed):
- Remove the added up special cancel
- Match up special's `hovering_accel_y` and `hovering_speed_y` to ROB's regular gravity/fallspeed (0.09/2.0)

stingers135:
> * Remove the added up special cancel, increase the ending gravity on ROB's normal cancel instead

What is robs normal cancel? Normally to exit the upb burner state, you have to do an action like aerial or airdodge

stingers135:
Changes made

SkewedAskew (approved):
Bair charge logic will need rewriting as I developed a cleaner way of doing it, but that can come in a followup PR

PR #1433, opened 2023-03-09
Teleport wallride logic improvement

jobrien97:
Teleport wallride behavior (to prevent bouncing off of sloped walls) now only applies on sloped walls, where it was previously being applied on flat walls as well.

This applies to:
- Palutena
- Mewtwo
- Zelda

Wallride logic is no longer run at all for Sheik, as she never had the bouncing issue in the first place.


SkewedAskew (approved):
Lgtm

heyha1400:
Hi, using 0.28.17-nightly the recovery of Palutena and Mewtwo while wall hugging the stage from under works (I assume correctly) and lets you recover and grab the ledge, but doing the same with Zelda results in her bouncing/getting send away from the stage/ledge at the top part.
If I understood it correctly, should Zelda not bounce away at the top, but behave the same as the teleport from Mewtwo and Palutena in that sense?

Tested on Battlefield and Final Destination.

PR #1439, opened 2023-03-11
Fix hanging on Salty Runback/Quick Exit inputs

### LINKED ISSUE #1347: Holding the Quick Exit input causes a slight hang
Quick Exit input can register for multiple frames in a row as long as you have the buttons held, for a maximum of 4 frames

This causes a slight hang right before exiting the match

### LINKED ISSUE #1348: Salty Rematch With ICs Runs Twice
Nana's ControlModule also inputs Salty Rematch 6f after Popo, causing the restart to happen twice, and taking noticeably longer.

jobrien97:
Fixes an issue where the game would hang for a second if the Salty Runback/Quick Exit inputs were held.
Nana will no longer perform Salty inputs 6f after Popo as well.

Fixes #1347 
Fixes #1348

### DIFF
--- a/fighters/common/src/opff/mod.rs
+++ b/fighters/common/src/opff/mod.rs
@@ -1,11 +1,4 @@
-use utils::{
-    *,
-    ext::*,
-    consts::*
-};
-use smash::app::{self, lua_bind::*, sv_system, sv_kinetic_energy};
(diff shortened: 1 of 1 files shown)


blu-dev (approved):
lgtm

PR #1442, opened 2023-03-13
Link Recovery Adjustments

MadVillainLordQ:
### Hero Spin:
- [!] No longer has fire properties when holding A during aerial version. (Up special bomb recovery is no longer possible)
- [-] Horizontal max aerial acceleration reduced


jobrien97:
Clip pls?

PR #1445, opened 2023-03-16
OPFF motion frame logic cleanup

jobrien97:
Adds `StatusModule::is_changing` checks to some OPFF for compatibility with new Smashline update --- namely, OPFF which use motion frame checks within status kind checks.

Also replaces motion frame usage in OPFF with status frames, where applicable.

Additionally, spacies' shine JCs come out on the proper frames now.


WuBoytH (approved):
E

PR #1446, opened 2023-03-17
Fix hitlag movement bug

jobrien97:
Fixes an issue where an attacker would not resume movement until 1 frame *after* hitlag was over.


SkewedAskew (approved):
Lgtm

PR #1462, opened 2023-03-19
feat: add 0/1 player smash and remove no contest screen

blu-dev:
If you pick time battle, all players set to random to nothing.

This also requires removing the no contest/result screen in time battles to not soft lock the game. I also removed the no contest in stock battle, it just goes back to CSS when you quit now.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -1,6 +1,8 @@
 #![deny(deprecated)]
 #![allow(unused)]
-#![allow(non_snake_case)]#![allow(unused_imports)]#![allow(unused_variables)]
+#![allow(non_snake_case)]
+#![allow(unused_imports)]
+#![allow(unused_variables)]
 #![feature(proc_macro_hygiene)]
 
 #[cfg(feature = "main_nro")]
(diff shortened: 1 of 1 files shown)


plugin/src/lib.rs
```diff
@@ -207,7 +216,9 @@ unsafe fn push_something(game_state: u64, amt: u32) {
     let mut val2 = *game_state.add(0x108 / 8);
     let mut val3 = if val2 - val1 != 0 {
         (val2 - val1) * 0x20 - 1
-    } else { 0 };
+    } else {
```
HisashiASM, line 219: Ew, don't like this, but it is whatever

plugin/src/lib.rs
```diff
@@ -294,17 +370,42 @@ pub extern "C" fn main() {
...
+            load_ingame_call_sequence_scene,
+            load_melee_scene,
+            game_end,
+            game_exit
+        );
     }
 
     #[cfg(not(feature = "runtime"))]
-    { utils::init(); }
+    {
```
HisashiASM, line 389: Seems to do it on one liner functions, but whatever (Does Rust support no curly brace on one line functions?)
`#[cfg(not(feature = "runtime"))]
          utils::init();`
or something?

HisashiASM (dismissed):
LGTM

plugin/src/lib.rs
blu-dev: don't think so bad

PR #1469, opened 2023-03-23
Ridley - fix downB aerial stalling bug

jobrien97:
Fixes an issue where Ridley could infinitely stall in the air if using downB out of sideB jump release.


SkewedAskew (approved):
Lgtm

PR #1470, opened 2023-03-23
Remove remaining held jabs

jobrien97:
DK, Peach, Dark Pit, and Shulk could still hold jab on whiff to repeat jab1 or complete their jab combo. This is now removed to match the rest of the cast.


MadVillainLordQ (commented):
LGTM

PR #1475, opened 2023-03-24
Nerf knockback smoke even more

jobrien97:
- Knockback speed threshold required to trigger knockback smoke increased 3.08 -> 4.0
- Knockback smoke farts were still occurring in some instances, so it now starts generating a few frames later into your kb arc than before


SuddyN (approved):
LGTM

PR #1477, opened 2023-03-25
Double air traction when above max jump speed

jobrien97:
Much like how on the ground, you undergo double traction while moving above your character's max walk speed, you will now undergo double air traction (air brake) when moving above your character's max jump speed in the air.

This is meant to balance the significant aerial momentum some characters gain when performing the recently implemented platdrop momentum transfer.


SuddyN (approved):
well-written

PR #1484, opened 2023-03-25
Sheik - fix jump cancel & reinstate airdodge cancel on Bouncing Fish return

jobrien97:
Previously, Sheik could only jump cancel Bouncing Fish return before or on frame 14.
Now, Sheik can both jump cancel or airdodge cancel *after* frame 14.


SkewedAskew (approved):
Oops

PR #1485, opened 2023-03-25
Lucas - add smoke gfx on dsmash2

jobrien97:
Makes 2nd dsmash more visually clear that it has been input.


HisashiASM (approved):
Meme! Approved!

PR #1486, opened 2023-03-25
Kirby - remove rising upB ledgesnap

jobrien97:
Kirby was unintentionally able to grab ledge during the first few rising frames of upB.


WuBoytH (commented):
Genuinely why does this hook exist
All of this should be handled in acmd

jobrien97:
> Genuinely why does this hook exist
> All of this should be handled in acmd

Not wrong

WuBoytH (approved):
based

PR #1489, opened 2023-03-27
Temporarily revert CSS/SSS switch and 0p smash

jobrien97:
Until bugs/crashes can be ironed out


HisashiASM (approved):
Approved, hopefully we can fix these soon.

PR #1496, opened 2023-03-27
Rosalina Down B tweaks

jaccob101:
Can now use vanilla gravity pull if you release the special button before 10 frames. Move is only locked out for 2 seconds of the 5 second warp cooldown. You still cannot warp unless you are in vision range of luma, within range, and have waited for the 5 second timer to completely run out which is indicated by a flash on the UI, Rosalina, and her Luma if alive. Each use adds 1.5 seconds back to the cooldown, so spamming it will keep your warp from refreshing.

### DIFF
--- a/dynamic/src/consts.rs
+++ b/dynamic/src/consts.rs
@@ -403,6 +403,7 @@ pub mod vars {
 
 			// flag
             pub const IS_TICO_DEAD: i32 = 0x0105;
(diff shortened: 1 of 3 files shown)


hdr-balance-backroom:
Currently we do not think Rosalina should have a way to negate projectiles while also possessing such strong defensive presence via Luma. As such, we will not be considering this pull request at this time

PR #1498, opened 2023-03-27
Greninja - fix sideB reappearance

jobrien97:
Fixes an issue where grounded sideB reappearance would send you offstage.


SkewedAskew (approved):
Lgtm

PR #1505, opened 2023-03-30
Little Mac - fix KO Punch

jobrien97:
Fixes an issue where grounded KO Punch would stick to the ground in the aerial animation.


SkewedAskew (approved):
Lgtm

PR #1509, opened 2023-03-31
Ice Climbers Recovery Nerf

SuddyN:
- Reduce side special initial vertical speed to vanilla
- Decrease horizontal drift


jobrien97 (changes requested):
air_stick_accel_mul_x should be placed before the other params, as that is the original order

SuddyN:
> air_stick_accel_mul_x should be placed before the other params, as that is the original order

corrected (and noted for my Ken changes, lol)

jobrien97 (approved):

PR #1515, opened 2023-04-02
Sora Blizzaga Rework + Changes

MyFriendAaron:
Capture

**Draft 8**:
Sonic Blade startup animation sped up 
- Sonic Blade 1 hitbox 17 -> 15
- Increased endlag 17 frames -> 21

Slows fall speed through the air.
Control Drift in the air based on stick position.

Hitbox Stats:
Frame: 10-15 active frames  
 Damage - 4.5
  Angle - 105
  KBG - 65
  FKB - 0
  BKB - 85
  
  25-30, 40-45, 55-60 , 70-75, 85-90
  Damage - 4.5
  Angle - 361
  KBG - 60
  FKB - 0
  BKB - 30

   
**Draft 7**:
Sonic Blade startup animation sped up 
- Sonic Blade 1 hitbox 17 -> 13

Blizzaga startup animation sped up
- 12 -> 10

**Draft 6**:
Re-implemented 3 hitboxes, to give the move disjoints to better match his current kit
Angle and knockback of moves all made uniform.
Damage of smaller outside hitboxes are 1.5 instead of 4.5 like the center hitbox still is.

KBG: 100 -> 65
BKB: 77 -> 70

New move effect:
image

New move hitboxes:
image

**Draft 5**:
Hitbox has been adjusted so there is only 1 large central hitbox
- KBG - 100
- Damage - 4.5
- Size - 8.25

Speed of Blizzaga has also doubled and can be jump canceled any time 2 frames after first hitbox comes out

New Blizzaga Hitbox Visualized:
image

**Also added to this change**:

Up special:
  Has been adjusted to have a far weaker sdi modifier to fall in line with Link's up special:
    - 1.0 -> 0.3
  Last hit of up special was found to be coming out far too late
  Activation of the last hitboxes of up special happen 3 frames earlier and hit boxes are active for 6 frames instead of only 3
  *This should hopefully help out with characters falling out of up b relatively frequently

Jab 2:
  Small adjustment made to allow Sora to finish jab combo

**Draft 4**: 
Updated KBG and BKB of center hitbox to more closely match that of Dark Pit
- KBG - 85
- BKB - 77

**Draft 3**:
Hit-boxes have been adjusted:
- Center hit-box is now ID 0 so it takes priority so jump cancel will feel fluid.
- Front and back hit-boxes made smaller and closer to Sora's model to balance out jump cancel addition

New hit-boxes:
image

New Center Hit-box Stats:
    - KBG - 120
    - FKB - 0
    - BKB - 90

**Draft 2**: 

Added ability to jump cancel if center hit-box is hit.

**Draft 1**:

Sora's blizzaga have been altered from a cone-shaped ice attack that would freeze an opponent to an AOE style combo-starter with 3 large hit-boxes.

The 3 hit-boxes are both in front and behind Sora, as well as around him to protect from stomp-style aerials.

The hit-boxes will flash 3 times in total, with 5 active frames each, with 2 frames of not hit-boxes to fall in-line with the blizzaga shown here:
https://youtu.be/mDWsAiXthac?t=306

Hit-boxes:
image

Hit-box Stats:
  Red:
    - Damage - 2.1
    - Angle - 105
    - KBG - 80
    - FKB - 70
    - BKB - 0
  Magenta:
    - Damage - 2.1
    - Angle - 105
    - KBG - 80
    - FKB - 0
    - BKB - 60
  Yellow:
    - Damage - 2.1
    - Angle - 75
    - KBG - 80
    - FKB - 70
    - BKB - 0
    
These stats are set the way they are, so that both the outside hit-boxes will pull the opponent towards the center hit-box, which will have knock-back weak enough that Sora may follow up with aerials, attempt a read, etc.

Here's a better look at the effect without the hit-boxes visualized:
image
*Keep in mind that they all spin in the same direction which is clockwise and the animations are de-synced

This is a first draft and I have not gotten time to see how this works outside of Training Mode, so I am super open to suggestions on making adjustments for this for balancing purposes and aesthetic adjustment. I simply aim to make the move deeper and less of a free edge-guard gimmick.


fighters/trail/src/opff.rs
```diff
@@ -114,6 +114,15 @@ unsafe fn magic_cancels(boma: &mut BattleObjectModuleAccessor) {
...
     }
+    // Blizzard jump cancel
+    if (boma.is_status(*FIGHTER_TRAIL_STATUS_KIND_SPECIAL_N2)
+        && AttackModule::is_infliction_status(boma, *COLLISION_KIND_MASK_HIT) 
+        && VarModule::get_int(boma.object(), vars::common::instance::LAST_ATTACK_HITBOX_ID) == 0) {
+                boma.check_jump_cancel(false);
+                AttackModule::clear_all(boma);
+                WorkModule::off_flag(boma,  *FIGHTER_TRAIL_INSTANCE_WORK_ID_FLAG_MAGIC_SELECT_FORBID);
+                WorkModule::on_flag(boma,  *FIGHTER_TRAIL_STATUS_SPECIAL_N2_FLAG_CHANGE_MAGIC);
+    }
```
SuddyN: please add a check to make sure the player is not currently in hitlag `!boma.is_in_hitlag()`

SuddyN: also, `AttackModule::clear_all(boma);` is redundant: hitboxes will be cleared upon changing status

SuddyN (changes requested):

fighters/trail/src/opff.rs
MyFriendAaron: Understood, have adjusted based off your feedback

SuddyN (dismissed):

SuddyN (dismissed):

PR #1535, opened 2023-04-10
Move parry from shield release to dedicated input

blu-dev:
# Summary of Changes
This pull request moves the ability to parry (aka perfect shield) from shield release to a dedicated input. Depending on the user's controller settings, that input is either Shield + Special or Shield + Taunt (with the alternative being a macro for shield locking).

This pull request also removes the macro for instant shield drop of Shield + Taunt.

The following character changes were made:
- Inkling's neutral and side specials can be cancelled during the first 5 frames with shield in order to transition to the charging ink status.
- Steve's ability to move crafting table with Shield + Special has been removed.
- Donkey Kong's barrel has temporarily been moved to Shield + Taunt by default (cannot be changed, even if taunt is picked to be the parry input) until more concrete DK adjustments are finalized

# Detailed Breakdown
## Parry Input
On the control settings, screen, when clicking the additional options in the bottom left corner, you will be presented with the additional options `Parry Input` and `To Be Decided`. `To Be Decided` is a placeholder setting that does nothing, although we suggest leaving that off for now.

* Slight correction, this image should say "Special" instead of "Shield"
image

## Shield Dropping
Previously, there were 3 ways to shield drop in HDR
* The Axe Method (due to spotdodge thresholds, this is slight inconsistent)
* Shield Lock + Stick flick down
* ~~Shield + Taunt~~ This has been removed.

Shield Locking is performed by pressing either two shield buttons or shield + special. Because the default parry input is now Shield + Special, the default macro has been moved to Shield + Taunt. If you change your parry input to taunt, Shield + Special will continue to be a shield lock macro.

## Parry Advantage, Window, and Whifflag
When you press shield special, you enter your parry animation with a 5f window to land your parry. During this window, standard HDR parry rules apply, which include not taking damage and extending hitstop for the move as well as reflecting projectiles.

If you do not connect with a hitbox during this time, the remainder of your animation will play out over the next 25 frames. You will be inactionable, this is known as the "parry whifflag".

If you DO connect hitbox, once hitlag is over, the attacker will be able to act and cancel out of the remainder of their parry animation.
Some caveats apply:
* The buffer is cleared on the first frame of successful parry so you do not get accidental neutral specials or special moves (or taunts if that is your option).
* You will not re-enter shield unless you release the button and re-press it. If you wait until the end of your parry animation, and are still holding shield, you will re-enter shield at that time. This is to allow for more consistent ways to parry multihits that doesn't require releasing and re-pressing the shield button over and over again.

This gives parry a increasingly more noticeable advantage stage when powerful moves are parried -- a distinct difference from how parry mechanics worked in HDR previously and in Ultimate.

## Effects
When you press your parry input, a slight metallic clanking sound will play as an audio queue, and a smaller version of the parry-hit effects will be played to visually indicate the transition to parry and allow an opponent to react. 

## Required Assets (Use Launcher)
[parry-layout-assets.zip](https://github.com/HDR-Development/HewDraw-Remix/files/11187484/parry-layout-assets.zip)


SkewedAskew (approved):
Needs less auto-formatting

PR #1544, opened 2023-04-12
Cloud Changes

jdipalermo27:
Take away the egregious alpha stuff (finishing touch, bair back swing, long limit timer, and unchanged vanilla up b 2) and in turn buffing some options and making other's more consistent.

<==Cloud==>

-BAIR
~[-] Active Frame Window 10-17 --> 10-14
-UPSMASH
~[+] Early and Middle Hitbox KBG: 40 --> 50
-Grounded Climb Hazard 1
 Initial Hit
~[R] Hitbox ID 2 Angle: 102 --> 80
~[R] KBG: 100 --> 5
~[R] FKB: 80 --> 0
~[R] BKB: 0 --> 100
 Rising Hit
~[R] Hitbox ID 0 Angle: 94 --> 50
~[R] Hitbox ID 1 Angle: 96 --> 50
~[R] KBG: 100 --> 5
~[R] FKB: 121 --> 0
~[R] BKB: 0 --> 100
-Climb Hazard 2
~[R] KBG: 100 --> 120
~[-] FKB: 110 --> 0
~[R] BKB: 0 --> 30
-Limit Cross Slash
 Final Hit
~[R] Angle: 50 --> 30
~[R] KBG: 151 --> 55
~[R] BKB: 30 --> 97
-Grounded Finishing Touch
~[-] BKB: 60 --> 51
-Aerial Finishing Touch
~[-] BKB: 60 --> 37
-Limit Timer
~[-] Timer: 45 sec --> 20 sec

### DIFF
--- a/fighters/common/src/function_hooks/edge_slipoffs.rs
+++ b/fighters/common/src/function_hooks/edge_slipoffs.rs
@@ -39,7 +39,8 @@ pub unsafe fn init_settings_edges(boma: &mut BattleObjectModuleAccessor, situati
             *FIGHTER_STATUS_KIND_GUARD_DAMAGE,
             // *FIGHTER_STATUS_KIND_ESCAPE_AIR,
             // *FIGHTER_STATUS_KIND_ESCAPE_AIR_SLIDE,
-            *FIGHTER_STATUS_KIND_DAMAGE].contains(&status_kind) {
+            *FIGHTER_STATUS_KIND_DAMAGE,
+            *FIGHTER_STATUS_KIND_AIR_LASSO_LANDING].contains(&status_kind) {
             fix = *GROUND_CORRECT_KIND_GROUND as u32;
         }
 


jdipalermo27:
Just to write the changes in layman's terms;
The bair change is to remove the back swing hitbox that really has no reason to be there, but is annoying when trying to punish it.
The climb hazard rising hit change is to make grounded climb hazard a consistent oos and anti air option, as the current iteration can be sdi'd and punished on hit.
The falling climb hazard change is to balance the spike hitbox on the move to be in line with other spikes in the game (it's current iteration will kill some characters at 0).
The finishing touch changes are to move the window on kill confirms up and to distinguish the knock back between grounded and aerial finishing touch. You can still confirm into grounded ft on platforms, so nerfing it was necessary to balance the move.
The limit cross slash buffs are intended to make the move more effective offstage with higher bkb and lower kbg, the move can be cced in it's current iteration, which makes for good counterplay around a move designed this way.
The upsmash buffs were another thing added to round out Cloud's kit more, the move in it's current form has very little use as it has low knockback, meaning it won't kill, and it also falters to up b as an oos and anti air option.
The bow on top of this patch is the reduction of limit timer to 20 seconds. My view of limit is that gaining a limit break should be a planned option, the way it is with 45 seconds makes it a condition to reach every stock, or an inevitability. With this change I hope to make Cloud think more to get these confirms, having to plan for when they get limit and give them a limited opportunity to expend it. I tried 15 seconds and it felt like it was just a bit too tight of a window, but that may be closer to what we're looking for. 20 seconds is a good medium for testing at the moment though and all of these can be adjusted over time to fully balance the character.

hdr-balance-backroom (changes requested):
- Revert Limit Cross Slash changes
- Lower angle on up tilt to decrease vortexing potential

PR #1545, opened 2023-04-12
Fix dash grab jostle

jobrien97:
Extra jostle on dash grab was not being applied

### DIFF
--- a/fighters/common/src/general_statuses/catch/catchdash.rs
+++ b/fighters/common/src/general_statuses/catch/catchdash.rs
@@ -31,7 +31,6 @@ unsafe fn status_end_CatchDash(fighter: &mut L2CFighterCommon) -> L2CValue {
 
 #[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_bind_address_call_status_end_CatchDash)]
 unsafe fn bind_address_call_status_end_CatchDash(fighter: &mut L2CFighterCommon, _agent: &mut L2CAgent) -> L2CValue {
-    JostleModule::set_overlap_rate_mul(fighter.module_accessor, 1.0);  // reset to 0.3 overlap rate
     fighter.status_end_CatchDash();
     0.into()
 }
--- a/fighters/common/src/general_statuses/catch/catchdash.rs
+++ b/fighters/common/src/general_statuses/catch/catchdash.rs
@@ -16,14 +16,15 @@ fn nro_hook(info: &skyline::nro::NroInfo) {
     }
 }
 
-#[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_status_pre_CatchDash)]
+#[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_status_pre_CatchDash_common)]
 unsafe fn status_pre_CatchDash(fighter: &mut L2CFighterCommon) -> L2CValue {
     JostleModule::set_overlap_rate_mul(fighter.module_accessor, 6.666);  // 0.3 (base overlap rate) * 6.666 = 2.0 overlap rate
     call_original!(fighter)
 }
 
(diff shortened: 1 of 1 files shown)


WuBoytH (commented):
You got schmixed by the bind_call_address version too, huh...

to be *double* sure, I would also hook the one you had before. In fact, I would call `fighter.status_end_CatchDash` *from* the `bind_call_address` version, to make *triple* sure

fighters/common/src/general_statuses/catch/catchdash.rs
```diff
@@ -16,14 +16,15 @@ fn nro_hook(info: &skyline::nro::NroInfo) {
...
     JostleModule::set_overlap_rate_mul(fighter.module_accessor, 6.666);  // 0.3 (base overlap rate) * 6.666 = 2.0 overlap rate
     call_original!(fighter)
 }
 
-#[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_status_end_CatchDash)]
-unsafe fn status_end_CatchDash(fighter: &mut L2CFighterCommon) -> L2CValue {
-    JostleModule::set_overlap_rate_mul(fighter.module_accessor, 1.0);  // reset to 0.3 overlap rate
-    call_original!(fighter)
+#[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_bind_address_call_status_end_CatchDash)]
+unsafe fn status_end_CatchDash(agent: &mut L2CAgent) -> L2CValue {
```
WuBoytH: oh yeah also this should be `status_end_CatchDash(fighter: &mut L2CFighterCommon, _agent: &mut L2CAgent) -> L2CValue`. I typically leave the L2CAgent alone.

WuBoytH (changes requested):

fighters/common/src/general_statuses/catch/catchdash.rs
jobrien97: done

fighters/common/src/general_statuses/catch/catchdash.rs
```diff
@@ -10,20 +10,28 @@ pub fn install() {
...
 #[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_status_end_CatchDash)]
 unsafe fn status_end_CatchDash(fighter: &mut L2CFighterCommon) -> L2CValue {
     JostleModule::set_overlap_rate_mul(fighter.module_accessor, 1.0);  // reset to 0.3 overlap rate
-    call_original!(fighter)
+    0.into()
+}
+
+#[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_bind_address_call_status_end_CatchDash)]
+unsafe fn bind_address_call_status_end_CatchDash(fighter: &mut L2CFighterCommon, _agent: &mut L2CAgent) -> L2CValue {
+    JostleModule::set_overlap_rate_mul(fighter.module_accessor, 1.0);  // reset to 0.3 overlap rate
```
WuBoytH: I mean if you're calling `fighter.status_end_CatchDash()` you wouldn't need this line here
otherwise it's fine

WuBoytH (changes requested):

fighters/common/src/general_statuses/catch/catchdash.rs
jobrien97: it's a safeguard, Mr. triple sure

WuBoytH: I mean it in the sense that if the bind_address version calls the normal one, and the normal one is modified, you'd be making sure every instance of it is being changed

jobrien97: oy vey

WuBoytH (approved):
you're welcome

PR #1546, opened 2023-04-13
Camera parameter standardization + camera adjustments

jobrien97:
This programmatically standardizes some of the camera's parameters to be the same *across all stages*, without needing to individually edit each stage's .stprm file.

The following parameters have been standardized:
- Minimum camera zoom
- Target interpolation rate (character tracking speed)
- Camera horizontal rotation rate
- Camera vertical rotation rate
- Unrestricted Camera parameters

Along with this, some changes have been made to the normal battle camera:
- Minimum camera zoom value has been increased from 100 -> 120, meaning the camera will be less zoomed in when characters are close together
- Target interpolation rate has been decreased from 1.0 -> 0.9
- Horizontal rotation rate has been decreased from 0.1 -> 0.0
- Vertical rotation rate has been decreased from 0.1 -> 0.0

*These 4 changes should alleviate some issues regarding camera disorientation.*

Minimum zoom before:

image

Minimum zoom after:

image

The new minimum zoom value should more closely match Melee/PM in how much space characters take up on screen, as you can see below:

minzoomcompare

To be merged with HDR-Development/hdr-private#332

### DIFF
--- a/fighters/common/src/function_hooks/camera.rs
+++ b/fighters/common/src/function_hooks/camera.rs
@@ -20,6 +20,32 @@ unsafe fn parse_stprm_normal_camera_min_distance(ctx: &mut skyline::hooks::Inlin
     }
 }
 
+// Standardizes swing_rate_y for all stages
+#[skyline::hook(offset = 0x2620cc8, inline)]
+unsafe fn parse_stprm_swing_rate_y(ctx: &mut skyline::hooks::InlineCtx) {
+    let swing_rate_y: f32 = 0.1;
+    asm!("fmov s0, w8", in("w8") swing_rate_y)
+}
+
+// Standardizes normal_camera_vertical_angle and normal_camera_look_down_vertical_angle for all stages
+#[skyline::hook(offset = 0x2620e54, inline)]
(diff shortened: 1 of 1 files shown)


fighters/common/src/function_hooks/camera.rs
```diff
@@ -9,8 +9,234 @@ unsafe fn normal_camera(ptr: u64, float: f32) {
...
+    else if hash == hash40("normal_camera_look_down_vertical_angle") {
+        if angle < -25.0 {
+            angle = -25.0;
+        }
+    }
+    asm!("fmov s0, w8", in("w8") angle)
+}
+
+// Standardizes target_interpolation_rate for all stages
+#[skyline::hook(offset = 0x2620fec, inline)]
```
blu-dev: There must be a better way to do this.

Are all of these hooks in the same function? Is it not possible to instead modify, at the end of the function, the location in memory they are stored?

blu-dev (changes requested):

blu-dev (approved):
lgtm

PR #1547, opened 2023-04-13
Little Mac - grounded Jolt Haymaker bugfix

SkewedAskew:
Fixes an issue where grounded Jolt Haymaker would immediately transition into the swing unless Special was only pressed for a single frame


jobrien97:
Bruh I thought this was the intended behavior lmao

jobrien97 (changes requested):
I know the easy solution to fixing these bugs is to throw it in the status exception list, but I ask that you try other solutions first. In this case, it seems you'd just need to `ControlModule::reset_trigger` on status change into `SPECIAL_S_JUMP`. I can PR this

SkewedAskew:
Fair enough, didn't know of that one

jobrien97 (approved):
well uh, i did the changes

PR #1550, opened 2023-04-13
Fix item tosses

### LINKED ISSUE #1540: Fix item tosses

jobrien97:
Fixes an issue where items would be released 1 frame too soon into your item toss animation, causing issues with item release points.

Fixes #1540


jobrien97:
@WuBoytH This is the only solution I am currently aware of other than...editing every item toss ACMD script in the game...

WuBoytH (approved):
I agree unfortunately

jobrien97:
> I agree unfortunately

Correction; items are now released on the *correct* frame. Before, item toss frames were defined w 1f delay in mind. So yeah

PR #1554, opened 2023-04-14
Fix roll/getup roll end momentum

jobrien97:
Fixes issues where some characters (most notably Mr. G&W) retained some momentum upon canceling the end of their roll or getup roll animation.


SkewedAskew (approved):
Lgtm

PR #1584, opened 2023-04-30
Palutena Teleport Hotfix

jaccob101:
Sweep formatting error broke grounded teleport. Now can jump out of grounded teleport again. In addition, the jump keep window on up b was listed as 3 frames but was only 2 frames in frame-by-frame. Changed to 3 to match what was written.


jobrien97 (approved):
lgtm

PR #1586, opened 2023-04-30
Pac-man recovery adjustments

jobrien97:
### SideB:
- Now sends into freefall
- Max distance reduced slightly
- Max damage increased 12% -> 16%
- Super armor removed

### UpB:
- Max air accel multiplier reduced 1.3 -> 1.0
- Max air speed multiplier reduced 1.4 -> 1.0
- Custom ledgegrab box implemented, lower than before


fighters/pacman/src/opff.rs
```diff
@@ -28,9 +28,20 @@ unsafe fn bonus_fruit_toss_ac(boma: &mut BattleObjectModuleAccessor, status_kind
     }
 }
 
+unsafe fn side_special_freefall(fighter: &mut L2CFighterCommon) {
```
WuBoytH, line 31: Why is this in opff
This could just as easily be done in acmd (and probably should be, if you don't want to bother with the status edit)

jobrien97: What? Why would this be in ACMD?

In the event we change its FAF for whatever reason it would break the logic in ACMD

jobrien97: I honestly dont see your point lol, this is a perfect use case for opff (not wanting to edit a convoluted status script)

WuBoytH: ... are you not completely removing its faf by making it go into special fall?

jobrien97: Sure, but the point still stands. Don't have to edit motion list this way/avoids changing status in ACMD, which should never be done

WuBoytH: Fair

WuBoytH (approved):

PR #1588, opened 2023-05-01
Lucario recovery adjustment

jobrien97:
### UpB:
- Startup increased 15f -> 25f (matches P+)
- UpB end gravity value increased 0.18 -> 0.3
  - Allows easier sweetspotting with upB cancels
- Ground bounce now always land cancels


SuddyN (commented):
lets slightly increase aerial USpecial startup
- makes it easier to pick a direction in time
- makes it easier for the opponent to interrupt

O1yyy:
^^^ please 200% this

PR #1602, opened 2023-05-06
Kazuya changes

jdipalermo27:
<==Kazuya==>
Parameters
~[+] Dash Speed: 1.775 --> 1.9
~[+] Fullhop Height: 18 --> 22
~[/] Fall Speed: .18 --> .16
~[-] Weight: 110 --> 104
FORWARD TILT
~[+] Jostle: 1 --> 6.666
~[-] Removed intangibility
UP TILT
~[-] Removed intangibility
DOWN TILT
~[+] Jostle: 1 --> 6.666
DASH ATTACK
~[+] Jostle: 1 --> 6.666
~[-] Removed intangibility
WGF
~[-] Removed excess shield stun multiplier
~[+] Jostle: 1 --> 6.666
EWGF
~[/] No longer pops out of shield
~[+] Shieldstun Multiplier: 1.2 --> 2.8
~[-] Shield Damage: 4 --> 0
~[+] SDI Multiplier: 1.0 --> 0.5
NAIR
Strong Hitbox (Arm)
~[+] Angle: 65 --> 270
~[+] BKB: 40 --> 7
~[-] Active Frames: 2 --> 1
Strong Hitbox (Outer Fist)
~[+] Angle: 65 --> 270
~[+] BKB: 40 --> 7
~[-] Active Frames: 2 --> 1
Weak Hitboxes
~[-] Active Frames: 2 --> 3
FAIR
Strong Hitbox (Toe)
~[+] Angle: 45 --> 40
~[+] KBG: 75 --> 85
Strong Hitbox (Foot)
~[+] Angle: 45 --> 50
Strong Hitbox (Inner Leg)
~[+] Angle: 45 --> 50
Strong Hitbox (Hip)
~[+] Angle: 45 --> 55
UPAIR
~[-] Removed intangibility
DEMON GOD FIST
~[+] Sdi Multiplier: 1.0 --> 0.5
JUMP SIDE KICK
~[+] BKB: 65 --> 80


jdipalermo27:
The reasoning behind these changes is to keep the feeling and flow of kazuya, being a tekken character, while altering his tools to better fit a rushdown/cqc character like falcon or little mac.
I have increased kazuya's initial dash speed, fall speed, fullhop height, and decreased his jump squat to frame 3 as a buff to his movement, in return I've lowered his weight to be in line with captain falcon's at 104, where previously he shared a weight with incineroar, a much slower cqc character.
This buff in movement also warranted the removal of intangibility on a large amount of his moves. While these are big nerfs, I do believe the movement changes are a net buff to his kit and his overall viability. While the intangibility and weight are big components of kazuya's strength, they aren't components that define him as a character.
My next change was to ewgf. Previously this move would pop people out of shield, which I believe is a hindrance of the move as it resets neutral and doesn't give kazuya the full ability to push his advantage beyond some more stage control. I instead have changed the move to hit on shield like a normal attack, but in this case electric will have a 2.8 shield stun multiplier, making it +3 on shield, a change which gives characters with a frame 5 jump squat the ability to jump away or wavedash away before kazuya can electric again, but it gives kazuya the ability to punish any defensive option the opponent goes for on a read and not a reaction. I have also halved the sdi multiplier on electric to allow for conversions out of the move, but to give opponents the opportunity to make his confirms much harder to land.
Another change was the increase of jostle to a lot of Kazuya's grounded options, this was to account for the fact that a lot of kazuya's moves previously would crossup point blank, a challenging disadvantage to work around for a character who is supposed to be in your face.
I made small changes to nair and fair in order to improve the feel of the moves. For nair I removed the weak hit from the initial hit as timing the move properly would sometimes not feel rewarding, I also reduced the activity on the initial hit by a frame and increased the weak lingering hit by a frame to keep the move precise to land.
For fair I changed the angle and knockback on the initial hits to give the move more impact. Previously I felt the initial hit and lingering hit felt very similar and very weak, not the initial hit will send more outward on the toe hitboxes and more upward on the hip hitboxes, this will lead to combos for the inner hitboxes at low and mid percents, tech chases for the outer hitboxes at mid percents, and edgeguards for the toe hitbox at higher percents.
The last change I made was to kazuya's back uptilt, the move was lacking much use as back tilt is already an intangible anti air with hurtbox shifting, meanwhile Jump Side Kick has a much smaller hitbox, so I decided to increase the bkb on it to shift it's kill window and reward players more for landing the move.

jdipalermo27:
Also please ignore the "Small Snake Changes" it seems I wrote the kazuya changes over my initial Snake Pr.

jdipalermo27:
I reverted jump squat to frame 5, initial fh accel, and the base knockback on nair.

PR #1618, opened 2023-05-11
Mario Firebrand Input Change

Nakedkoops:
Mario firebrand input is now B + A rather than holding B to make it more consistent to input.


SuddyN:
B+A seems less accessible than held B, what's the reasoning behind this change?

Nakedkoops:
Honestly, I might just delete this pr, I thought considered this change because it can help with distinguishing fireball and firebrand control wise. However, after talking about it in the mario chat, this change might not be the best, and even then most of us might have our muscle memory screwed up.

PR #1632, opened 2023-05-25
Megaman Minor Buffs

jaccob101:
This is meant to be a quick hold-over buff. not meant to interfere with alleged rework ideas posted in the megaman channel. 

Back Air:
~ [+] KBG: 183 -> 200
Kill percent from ~180% on inkling with di up/in on fd to %142, makes killing less miserable without making gimps absurd hopefully. 

Forward Air:
~ [+] BKB: Sweetspot 53 -> 55
~ [+] KBG: Sweetspot 89 -> 91
Overall allows it to kill earlier on the ground without making it kill absurdly early when juggling or combo'd into. 

Lemons: 
~ [+] Lifetime: 17 -> 32 frames

Lemons have range and no flinch, leaving it lackluster. This range increase allows megaman to more safely use it and take up space
Distance Comparison:
Before:
beta lemon
After:
revamp megaman pr pellet range

### DIFF
--- a/fighters/rockman/src/acmd/aerials.rs
+++ b/fighters/rockman/src/acmd/aerials.rs
@@ -37,9 +37,9 @@ unsafe fn rockman_attack_air_f_game(fighter: &mut L2CAgentBase) {
     }
     frame(lua_state, 8.0);
     if is_excute(fighter) {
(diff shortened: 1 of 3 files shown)


hdr-balance-backroom (dismissed):
- Increasing the lifetime of pellets is fine, in fact do it by 2 more frames than proposed (to 25)

hdr-balance-backroom:
> * Increasing the lifetime of pellets is fine, in fact do it by 2 more frames than proposed (to 25)

Actually you can even go to 32

MadVillainLordQ (approved):
LGTM

PR #1640, opened 2023-05-31
Manual parry aesthetic adjustments

jobrien97:
### Animation:
Rather than using the successful parry animation, your character will enter a "bracing" pose for the parry window (5f), then will slowly return to idle during their parry whifflag (25f).

### GFX:
The screen will no longer shake from a parry attempt. Rather than using a smaller version of the successful parry GFX, your character will now emit a blue flash of light, which matches your parry activation bubble in size (the parry bubble was previously invisible).

### SFX:
A parry attempt will now use a quicker, quieter, and less abrasive sound effect.

Demonstration:

https://github.com/HDR-Development/HewDraw-Remix/assets/47401664/36ef483f-9749-4a86-919e-cd8d9cbf4a7d


blu-dev (dismissed):
LGTM

PR #1643, opened 2023-06-01
Dash attack input improvements

### LINKED ISSUE #1081: Cstick Down and tilt stick up don't do dash attack during initial dash, and don't register until late into initial dash
When inputting cstick down with tilt/smash stick or cstick up with tilt stick during initial dash, you will do those respective moves instead of dash attack even when holding forward the entire time. The input is also not registered until doing these inputs near the end of initial dash. This is not the case for c-stick left or right, which can do dash attack from the very first frame of initial dash.

As far as I'm aware, no other smash game has this issue and it makes doing moves like DACUS and DACDS a little awkward for people used to that input method from the older brawl days (a well known input for DACUS was c-stick down + up Z)

TL;DR:
- down tilt, down smash, and up tilt when done with c-stick do not do dash attack when holding forward during initial dash
- these moves' inputs also do not get accepted at beginning of initial dash

Replication: Attempt to do down tilt, down smash, or up tilt with the c stick right at the beginning of initial dash(to see the input not get read), and then near the end of initial dash(to see it do the move instead of dash attack)

jobrien97:
If holding forward during dash, a C-stick input in any direction can now trigger dash attack.

Fixes #1081

### DIFF
--- a/fighters/common/src/general_statuses/dash.rs
+++ b/fighters/common/src/general_statuses/dash.rs
@@ -355,7 +355,7 @@ unsafe extern "C" fn status_dash_main_common(fighter: &mut L2CFighterCommon, arg
         fighter,
         FIGHTER_STATUS_KIND_ATTACK_DASH,
         true,
-        WorkModule::is_enable_transition_term(fighter.module_accessor, *FIGHTER_STATUS_TRANSITION_TERM_ID_CONT_ATTACK_DASH) &&
+        fighter.left_stick_x() * PostureModule::lr(fighter.module_accessor) > 0.0 &&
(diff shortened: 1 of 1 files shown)


WuBoytH (changes requested):
You can't just remove the is_enable_transition_term check, otherwise it'll allow things like Mega Man Dash Attacking during Leaf Shield, Pyra Dash Attacking while Blazing End is out, etc...

jobrien97:
> You can't just remove the is_enable_transition_term check, otherwise it'll allow things like Mega Man Dash Attacking during Leaf Shield, Pyra Dash Attacking while Blazing End is out, etc...

dam u rite

WuBoytH (approved):
Someone else should def test to make sure it works but this looks fine to me

PR #1653, opened 2023-06-05
Install stage api callbacks only if TourneyConfig is valid

Coolsonickirby:
Will only install the API callbacks if the TourneyConfig is valid (allows stage-alts to be used by people who don't want to use the tourney stuff)


Angstdiscord:
Tested extensively, both with tournament mode on and off. LGTM

techyCoder81 (approved):
Looks reasonable to me

PR #1654, opened 2023-06-05
MinMin HotFix

CSharpM7:
Current fixes:
`Fixed an issue where Minmin would hop after completing an arms attack in the air`

TODO:
`Figure out what causes arms superslide`


SkewedAskew (approved):

jobrien97:
Is this supposed to fix the triple jump bug? It didn't if so

jobrien97:
Side note, ground-to-air jump momentum isn't working for Min Min rn. I can try to narrow it down but just want it to be documented

PR #1655, opened 2023-06-05
New Lucas down throw effects + miscellaneous FX fixes

oasis-sharp:
https://github.com/HDR-Development/HewDraw-Remix/assets/64927775/56a099e4-bb58-4505-af7c-737ca3a0db54

Lucas' down throw in HDR still makes use of Ultimate's bury effect, which in its sped up form is quite jarring visually. This PR changes the down throw to resemble his down throw in PM/Brawl, without visual the bury effect, and a new visual on the hitbox that makes it look like a PSI blast hits them out of the ground after they're initially hit. It matches up a bit better overall with his kit and looks better in combos and strings, and also makes more sense because what was that bury throw anyway.

The FaF of the throw should be the exact same. Similarly, the timing for the throw should be extremely similar (for combo DI), so this is neither a buff nor a nerf, purely a visual change

I've also fixed the timing of his voice lines on his side special to come out upon using the move rather than on the first frame, and made his down smash sound effect louder to better fit the strength of the move as previously it felt quite underwhelming to hit.

### DIFF
--- a/fighters/lucas/src/acmd/smashes.rs
+++ b/fighters/lucas/src/acmd/smashes.rs
@@ -420,7 +420,7 @@ unsafe fn lucas_attack_lw4_game(fighter: &mut L2CAgentBase) {
         }
         frame(lua_state, 20.0);
         if is_excute(fighter) {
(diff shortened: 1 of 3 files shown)


SkewedAskew:
Not being considered, we prefer the current animation

PR #1658, opened 2023-06-07
Fix smash stick up out of dash

jobrien97:
Smash stick up out of dash will usmash again, instead of dash attack.


SuddyN:
doesn't seem to work. tested online

SuddyN:
works now

SuddyN (approved):

PR #1660, opened 2023-06-07
Cloud - usmash hitlag reversion

jobrien97:
Reverts a previous change that lowered the hitlag multipliers on Cloud's usmash, making it feel weak.


SkewedAskew (approved):
Needs more comments

PR #1661, opened 2023-06-08
Rosalina teleport adjustment

SkewedAskew:
Prevents Rosalina from teleporting if Luma is put into hitstun before the two swap positions. If unsuccessful, Rosalina does not enter teleport cooldown


jobrien97 (dismissed):
lgtm

jobrien97 (approved):

PR #1678, opened 2023-06-12
Manual parry bugfixes/refinements

jobrien97:
- Fixed an issue where Yoshi's regular shield release would sometimes parry
- Fixed an issue where shield damage didn't have controller rumble
- Fixed a crash when entering the controls settings UI after previously changing settings and then playing a match
- Fixed an issue where taunt parry was only being applied to player 1
- Added 1f forgiveness window on both taunt and neutral special
  - Pressing shield on or before the second frame of taunt or neutral special will now activate Parry


UmbreonDiaz:
* Fixed an issue where taunt parry was only being applied to player 1
fix is confirmed working
* Fixed a crash when entering the controls settings UI after previously changing settings and then playing a match
fix is confirmed working
* Added 1f forgiveness window on both taunt and neutral special
* Pressing shield on or before the second frame of taunt or neutral special will now activate Parry
change works 
* Fixed an issue where shield damage didn't have controller rumble
fix is confirmed working

UmbreonDiaz:
* Fixed an issue where Yoshi's regular shield release would sometimes parry
bug fixed confirmed working

WuBoytH (approved):

PR #1680, opened 2023-06-12
MinMin fix duplicating double jump effect during arms attack

CSharpM7:
Main status of `AttackJumpAerial` initially changing motion, and then changing partial motions of MinMin seems to produce duplicate double jump effects. This guts the effect amcd for her leg's attack jump aerial motions, and calls the effect once during main to mitigate this


SkewedAskew (changes requested):
Min Min can jump on frame 1 of inputting an arm aerial, not sure if that was the case before. When jumping before the arm is released, no air jump ripple is shown

SkewedAskew (approved):

PR #1681, opened 2023-06-12
Fix DACUS disabling up/down smash during cancellable recovery frames

CSharpM7:
Dash Attack normally disables being able to transition into up/down smash outside the DACUS window. This also results in not being able to do up/down smash during the cancellable recovery frames of dash attack (tilts, jabs, and all other actions can still properly cancel


jobrien97 (approved):
good catch

PR #1682, opened 2023-06-12
Pyra - remove added hitstun on specials

jobrien97:
Removes the 10 frames of added hitstun on upB rise & dive, and more importantly, the added hitstun ranging from 5-40 (!!!) frames on sideB.


hdr-balance-backroom (dismissed):
- Motion rate the end of the sword spin to increase the time it takes to return to Pyra

SkewedAskew (approved):
Why does this have so many hits

PR #1684, opened 2023-06-13
Robin Thunder Rework

jdipalermo27:
Robin's thunder is an essential component of their kit, but it feels as though elthunder is the best option in most situations. Thunder is fast, but lacks range; thoron has good base knockback and damage, but lacks overall knockback and damage when compared to the other chargeshots; arcthunder is a move. It would be simple to just buff these options to make them more desirable, but it may become too strong, too fast. This proposed thunder rework, which has been discussed for awhile now in the robin chat, aims to add thunder to the vision of their kit, a strong and versatile character that requires proper resource management to take full advantage of their tools.
First change is to the thunder tome and the expense of each option. There will now be 8 charges to thunder tome, thunder will use up 1 charge or 1/8 of the tome, elthunder will use 2 charges or 1/4 of the tome, arcthunder will use 4 charges or 1/2 of the tome, and thoron will use the whole tome. This is a nerf to the amount of uses Robin gets of thunder per tome, as prior they could use it very liberally, now they will need to put more thought into each use of thunder as using up the tome will occur much more often, which not only loses the use of thunder as a projectile, but also as one of his main microspacing and movement tools in thunder cancelling. However, this justifies the buffing of Robin's thunder based projectiles.
Thoron has the most changes of the 3 projectiles, it now will use remaining charge as the basis of it's damage, with a full tome being the most powerful and damaging charge shot in the game and 1 charge being a small beam that maybe sets up for a tech chase and does around 6 damage.
Thunder has a lot of utility has a shut down option due to how fast it is, but the range is does this at pales in comparison to other projectile spacing tools (i.e. falco laser, lucas pkfreeze, ect.). The obvious answer is to increase the time it travels for, allowing it to punish opponents pinging him from across a platform and giving him the chance to take space.
Arcthunder, in it's current form, is underwhelming at best as the slow moving projectile is too slow to supersede robin's aerials for a proper punish, but too fast for robin to followup on. On top of that, the move is sdi'able during the multihits, which invalidates any stage control or frame advantage robin may gain from using the option. I've changed the move to have half an sdi multiplier of .5 to make the process of escaping it on hit much harder and I've increased the activity of the explosion hitbox to 24 frames from the previous 16, which makes the option +4 to +6 on shield at point blank and allows it to do 4.8 more points of damage, on top of the 8 frames of advantage more that robin gets to followup on hit.

Credit to jaccob101 for the thoron damage scaling code and being very helpful throughout this whole pr.


jdipalermo27:
<==ROBIN==>
Parameters
~[-] Thunder Tome Total Points: 18 --> 8
~[/] Elthunder Consume Points: 3 --> 2
~[/] Arc Thunder Consume Points: 5 --> 4
Thunder
~[+] Lifetime Frames: 15 --> 20
Arcthunder
~[+] SDI Multiplier: 1.0 --> 0.5
~[+] Explosion Active Frames: 16 --> 24
~[+] Rehit: 5 --> 4
Thoron
~[+] Damage: 3.3 --> 1 to 5, Thoron damage now scales with remaining charge.
~[-] Knockback Growth: 191 --> 145

hdr-balance-backroom (dismissed):
- Revert grab range increase

SkewedAskew (approved):

PR #1687, opened 2023-06-17
Sonic Side Special - Spin Boost

WuBoytH:
# Sonic Side Boost Rework - Spin Boost

Yes it's finally here.

https://github.com/HDR-Development/HewDraw-Remix/assets/26860994/764a642a-fc58-4328-b93f-ec76f4dd4b1b

https://github.com/HDR-Development/HewDraw-Remix/assets/26860994/b5ddeb36-eda2-4112-81a5-8abaa49da1b1

## Changelog

* Replaces Spin Dash with Spin Boost.
* Has a weak 4% hitbox on frame ~~7~~ 9 and an even weaker 1% hitbox on frame ~~12~~ 16.
* Grounded FAF 38, Aerial FAF 42
* Tapping Special keeps Sonic's vertical momentum and has a blue trail.
* Holding Special has Sonic hop into the air and has a yellow trail.
* If Spin Boost stays grounded, Sonic can edge-cancel after he starts moving and jump cancel (on hit only) after frame 12.
* If Sonic lands during Spin Boost, it land cancels with 15 frames of landing lag.
* If Sonic uses either Spin Boost or Spin Charge in the air, he cannot use both moves until he touches the ground or grabs a ledge. Spin Charge rolling off of a ledge counts as using it in the air.

* In addition:
  * Fixes an issue where Jab 3's built-in movement did not export properly, causing it to consistently whiff.
  * Allows turning around when starting Down Special (Spin Charge).

Includes assets for updated tips.

[include-assets.zip](https://github.com/HDR-Development/HewDraw-Remix/files/11780019/include-assets.zip)


hdr-balance-backroom (dismissed):
- Increase side b startup to frame 9
- Using side b should disable down b in the same airtime and vice versa

SkewedAskew (approved):

PR #1690, opened 2023-06-18
Ken Followup Adjustments

SuddyN:
## General
- Super1/Super2 inputs swapped

Super1 felt too easy to accidentally input in neutral. Super1 is the cheapest of the two supers, so it was not uncommon to accidentally spend it while saving up for Super2. Also, Ryu was planned to have a similar convention of super inputs (double quarter circle for the cheaper Super), so this will bring parity to the two of them when Ryu's changes are finished.

## Super1 (Shippu Jinraikyaku)
Final hit:
- angle 85 --> 75
- kbg 50 --> 95
- bkb 120 --> 85

Should make the move feel more rewarding at high percents, while still not killing any earlier than a non-command-input air heavy shoryu (which is to say there are still better, cheaper ways to kill than Super1).

## Super2 (Shinryuken)
- multihit SDI mul 1.0 --> 0.0
- launcher SDI mul 0.8 --> 1.0

Considering this super costs 5 bars (300% damage), it felt ridiculous to save up for the whole match and have it completely beaten by ASDI away. 0x multiplier is extreme, but necessary for a move that can only really be used once per match at the cost of worse neutral and advantage state the rest of the match.

## BThrow
- kbg 40 --> 64

Now similar in strength to Charizard FThrow, so not quite a kill throw, but can be strong against fast-fallers and punishing on bad DI (you have 41 frames to react lol).

## Heavy UTilt
- arms, head, and right leg are now invulnerable f8-f12 (duration of the first kick)

This restores some of the anti-air strength of Ken's old UTilt, which had similar invulnerability. That move also had an invulnerable upper body, but the upper body is not invulnerable in this change (only limbs and head)

## Heavy DTilt
- startup f6 --> f7
- fixed a bug where Heavy and Light DTilt hitboxes could both happen
- sweet/sour spots removed, now only one set of data
- damage 7.0/5.5 --> 6.0
- kbg 21 --> 53
- bkb 65 --> 55

The sourspot was a tipper intended to prevent the opponent from traveling too far from Ken. Now, the entire move uses one set of data, which feels more consistent while giving up that tipper.

This change prevents Ken from comboing DTilt into roundhouse kicks or shoryu at kill percents, and instead incentivizes him to make use of DSpecial to close the gap for a finishing move. In effect, Ken has to work harder to react to the opponent's DI and fall speed, but has the potential for better rewards if not mixed up by the opponent.

At lower and mid percents, Ken used to be able to chain DTilt and DSpecial together across the stage, which could only be broken by CC or ASDI down, but was particularly good against floatier characters. Because of higher KBG, this should be less of a problem.

## Proximity FTilt
- fixed an issue where meter was not incremented by landing the move


hdr-balance-backroom (dismissed):
Revert light utilt and dtilt hitlag

SkewedAskew (approved):

PR #1700, opened 2023-06-23
Fix amsah tech momentum + double traction refactor

jobrien97:
Amsah teching at high %s wouldn't properly transfer the remaining knockback momentum into the air when slipping off an edge. Most of the time you could simply grab ledge once slipping off an edge after your tech animation.

It now works as it does in Melee/PM.

Demonstration:

https://github.com/HDR-Development/HewDraw-Remix/assets/47401664/cbd5f1b4-9c1d-492f-be83-12efbc89b52f

This PR also moves double traction logic out of OPFF and into the proper energy functions.


blu-dev (approved):
Lgtm

PR #1702, opened 2023-06-23
Prevent jostle edge slipoffs

jobrien97:
This change prevents jostle from being able to push you off of edges, which may have been a major contributing factor to the complaints of "slipperiness" of our engine.

### Demonstrations:
(This would have previously pushed Falcon off the edge):

https://github.com/HDR-Development/HewDraw-Remix/assets/47401664/4da4b2b0-524a-46b2-b856-946090e26652

(This would have previously pushed DK off):

https://github.com/HDR-Development/HewDraw-Remix/assets/47401664/e048ce56-7443-478b-b715-3cb9e8382219

For reference, in Melee, jostle cannot push you off of edges *only* if you're in idle. In P+, jostle acts as it does in Ult and can push you off at all times. Here, jostle cannot push you off in any situation *except* knockdown (to allow pratfall combos to still exist).


WuBoytH (approved):
Needs more comments?

PR #1707, opened 2023-06-25
Fix DJC character jump aerial momentum

jobrien97:
Fixes an issue where DJC characters' (Ness, Lucas, Mewtwo, Peach, Sora, Yoshi) aerial attacks wouldn't retain momentum out of a running jump.


SkewedAskew (dismissed):
Sora's uair/dair/bair seem to still lose some momentum

jobrien97:
> Sora's uair/dair/bair seem to still lose some momentum

Fixed

WuBoytH (approved):
Needs less comments.

PR #1714, opened 2023-06-27
Reduce screenshake intensity

jobrien97:
Globally replaces screenshake strength levels as such:
Medium -> Small
Large -> Medium
Extra Large -> Large
Extra Extra Large -> Extra Large

Should make for less visual clutter, especially during doubles.


SkewedAskew (approved):
Seems good to me

PR #1720, opened 2023-06-30
Krool Blunderbuss bugfix

SkewedAskew:
Reverts crown catch behavior to how it was previously (1F FAF and auto-catch) and by extent fixes an issue where Blunderbuss would immediately enter the "spit" animation when trying to inhale the Kannonball


WuBoytH (approved):
Rip my code but yeah...

PR #1727, opened 2023-07-01
Lucas Up-Air FAF Fix

jaccob101:
When I gave feedback on the recent Lucas changes I accidentally calculated for faf 27 instead of 28 on up air, I was trying to match PM.


WuBoytH:
Ness Electric Magnet hit effect doesn't seem particularly necessary right now, save it for when someone actually adjusts Ness

SuddyN (approved):

PR #1731, opened 2023-07-01
allow jc grab oos if still holding grab

techyCoder81:
This PR should allow players to JC grab oos if they are still holding grab (button only) during jumpsquat. Currently, you cannot JC grab oos at all, because JC grab is disabled if jumping oos to allow buffered aerials oos. This is a minor improvement on that. Needs testing.

### DIFF
--- a/fighters/common/src/general_statuses/jumpsquat.rs
+++ b/fighters/common/src/general_statuses/jumpsquat.rs
@@ -136,8 +136,12 @@ unsafe fn status_JumpSquat_Main(fighter: &mut L2CFighterCommon) -> L2CValue {
         );
         return 0.into();
     }
+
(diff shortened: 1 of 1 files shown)


WuBoytH:
I don't really understand the point of this, why wouldn't you just shield grab at that point

techyCoder81:
The point is mostly just that it should be possible. In my own personal case, I happened to have built muscle memory to perform all grabs with JC grab lol. There's no reason why it shouldn't work here as well (right now, you get an usmash instead, even though what you input was grab).

SuddyN (commented):
I just keep crashing trying to test this. Unrelated, won't this make it more difficult for most players to aerial OOS, especially on characters with short jumpsquats? From my understanding, they would need to release the shield button within the length of their jumpsquat, or a jc grab would trigger. Could we check if shield was released during jumpsquat, and then re-applied for the JC grab?

SuddyN:
rejected

PR #1736, opened 2023-07-02
Adds Motion Lists in YML format to the main repo

WuBoytH:
Adds motion lists to the main repo in `romfs/source/` as motion_list.yml files. They will be automatically converted to motion_list.bin when building HDR, and *should* be included in the `hdr` mod folder from now on.

MUST NOT BE MERGED before https://github.com/HDR-Development/hdr-private/pull/396


WuBoytH:
Close to be reopened later

PR #1740, opened 2023-07-03
Float Hotfix (pt 2)

jaccob101:
I spent more time working on floats until I could fix a few more annoying issues. Fixes:
- Holding up no longer starts float on its own, now requires a jump input. Fixes a ton of misinputs that could happen when sdi'ing or manually inputting aerials, helps tap jump users
- Floats starting frame 7, not able to aerial out of float until frame 8 (robin)
- Aerials' float lockout wouldn't take into account cancel frames, not allowing you to float when actionable (made jump+down input eat djs)
- Appears to fix robin's grounded float instantly cancelling into fastfall if input too fast, likely due to the above changes to the aerial float restrictions.
- Robin float now brings up the meter every other tick, before using an aerial would prevent you from seeing the meter at all until the next time you used a wind move

### DIFF
--- a/fighters/common/src/opff/floats.rs
+++ b/fighters/common/src/opff/floats.rs
@@ -109,7 +109,7 @@ pub unsafe fn extra_floats(fighter: &mut L2CFighterCommon, boma: &mut BattleObje
                     MotionModule::change_motion(boma, Hash40::new("fall"), 0.0, 1.0, false, 0.0, false, false);
                 } else if status_kind == *FIGHTER_STATUS_KIND_JUMP {
                     if StatusModule::is_changing(boma) { //peach ground-float mechanic
-                        let pos = smash::phx::Vector3f { x: PostureModule::pos_x(boma), y: PostureModule::pos_y(boma) + 3.5, z: 0.0 };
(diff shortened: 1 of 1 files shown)


MadVillainLordQ (dismissed):
LGTM

fighters/common/src/opff/floats.rs
```diff
@@ -241,6 +242,10 @@ pub unsafe fn float_effects(fighter: &mut L2CFighterCommon, boma: &mut BattleObj
                 else if timer % 15 == 0 {  // every 15 frames
                     // drain 1 bar of Elwind
                     WorkModule::dec_int(boma, *FIGHTER_REFLET_INSTANCE_WORK_ID_INT_SPECIAL_HI_CURRENT_POINT);
+                    if timer % 30 == 0 {
```
jobrien97: This should be in an `else if` block outside of this `else if` block.

jobrien97 (changes requested):

fighters/common/src/opff/floats.rs
jaccob101: wouldn't that make it not run bc if it's divisible by 30 w/ no remainder it should always be divisible by 15 w/ no remainder right? or does code not work like that

jobrien97: my b, not `else if`, just `if`. I'd rather the logic be contained in its own block

but ya, ur right

jobrien97 (approved):

PR #1746, opened 2023-07-04
Parry input adjustment

jobrien97:
With the initial implementation, in order to trigger a Parry attempt, you could hold Special/Taunt as long as you wanted, then simply tap Shield to trigger it. There was no limit to how long Special/Taunt could be held before inputting Shield.

This interfered with several important character mechanics, such as Snake's grenade pull shield cancel, Bayo's gun charge shield cancel, and several other shield cancels.
This would also interfere with the Bidou control scheme.

This change makes it so the Special/Taunt input must be pressed **within the tap buffer window (5f)** to register a valid Parry input. The Shield button can still be held as long as you want prior to inputting Special/Taunt, but the inverse is no longer true.

This consequently allows Parry to now be bufferable, *as long as Shield + Special/Taunt are held at the same time once actionable*.


SkewedAskew (approved):
Yeah

PR #1754, opened 2023-07-06
init ledge_pos to NaN

SuddyN:
fixes Paper Mario right ledge tether trump bug, and prevents similar occurrences

too tired to test this personally


SuddyN:
someone should fix this up and do it properly but it's not something I'm interested in doing right at this time. Will revisit later.

PR #1756, opened 2023-07-06
Allows all grounded attacks to slide by default

jobrien97:
Right now, some characters' grounded attacks do not carry sliding momentum out of dash/run or DACUS/DACDS. Notable examples include Captain Falcon, Mewtwo, and Kazuya.

This allows all characters' jabs/tilts/smashes to be able to slide by default.


WuBoytH (approved):
bringus

PR #1764, opened 2023-07-07
Ganondorf fair visual effects

Nakedkoops:
Self explanatory, no changes to hitboxes or damage, just visual effects.

https://github.com/HDR-Development/HewDraw-Remix/assets/115908645/2b626d5b-3799-417b-97e0-4fb9cf77164d


CSharpM7:
I personally think we should stray away from any flashes during aerials that don't spike, namely for consistency.

Nakedkoops:
Umm, the flash is not part of the effects edits. That is (I think) base-ultimate. I could remove it if you want but again I didn't implement the flash.

SkewedAskew:
Not being considered, we prefer the current visuals

PR #1662, opened 2023-06-08
Robin Bug Fix: Tilt inputs correctly use bronze sword

jaccob101:
Before using the smash button would enable levin as expected, but tilt inputs wouldn't enable bronze sword. This makes it function logically.

PR #1765, opened 2023-07-07
Robin Levin Buffer/Switch

jaccob101:
(follow up on https://github.com/HDR-Development/HewDraw-Remix/pull/1662)
Previously I added a fix to make tilt input aerials switch to bronze. This had the unintended side-effect of lowering accessibility, as you needed a dedicated smash button or a frame perfect A+B Smash input to do levin nair. This adds 5 frames (match normal buffer frames) to input B or Smash at the start of each aerial. This should make Robin playable on more control schemes. LMK if frame buffer needs changing. I tested it to make sure that it uses and checks for durability and allows robin to throw the sword on resource burnout.

### DIFF
--- a/fighters/robin/src/opff.rs
+++ b/fighters/robin/src/opff.rs
@@ -32,6 +32,26 @@ unsafe fn elwind1_burn(fighter: &mut L2CFighterCommon) {
     }
 }
 
+unsafe fn levin_leniency(fighter: &mut L2CFighterCommon, boma: &mut BattleObjectModuleAccessor) {
+    let levin = *FIGHTER_REFLET_INSTANCE_WORK_ID_INT_THUNDER_SWORD_CURRENT_POINT;
+    if boma.is_status(*FIGHTER_STATUS_KIND_ATTACK_AIR) 
+    && boma.status_frame() <= 5 
+    && !fighter.is_flag(*FIGHTER_REFLET_INSTANCE_WORK_ID_FLAG_THUNDER_SWORD_ON) 
+    && boma.is_button_on(Buttons::Smash | Buttons::SpecialRaw) 
+    && !StatusModule::is_changing(boma)
+    && WorkModule::get_int(boma, levin) > 0 {
+        VisibilityModule::set_int64(boma, Hash40::new("sword").hash as i64, Hash40::new("sword_thunder").hash as i64);
+        fighter.on_flag(*FIGHTER_REFLET_INSTANCE_WORK_ID_FLAG_THUNDER_SWORD_ON);
+        VisibilityModule::set_int64(boma, Hash40::new("sword").hash as i64, Hash40::new("sword_thunder").hash as i64);
+        if WorkModule::get_int(boma, levin) == 1 {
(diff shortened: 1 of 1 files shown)


fighters/robin/src/opff.rs
```diff
@@ -32,6 +32,26 @@ unsafe fn elwind1_burn(fighter: &mut L2CFighterCommon) {
     }
 }
 
+unsafe fn levin_leniency(fighter: &mut L2CFighterCommon, boma: &mut BattleObjectModuleAccessor) {
+    let levin = *FIGHTER_REFLET_INSTANCE_WORK_ID_INT_THUNDER_SWORD_CURRENT_POINT;
```
SkewedAskew: This line should be placed inside the if block so it doesn't get run every frame (move the check for the point total inside it as well)

fighters/robin/src/opff.rs
```diff
@@ -32,6 +32,26 @@ unsafe fn elwind1_burn(fighter: &mut L2CFighterCommon) {
...
+    if boma.is_status(*FIGHTER_STATUS_KIND_ATTACK_AIR) 
+    && boma.status_frame() <= 5 
+    && !fighter.is_flag(*FIGHTER_REFLET_INSTANCE_WORK_ID_FLAG_THUNDER_SWORD_ON) 
+    && boma.is_button_on(Buttons::Smash | Buttons::SpecialRaw) 
+    && !StatusModule::is_changing(boma)
+    && WorkModule::get_int(boma, levin) > 0 {
+        VisibilityModule::set_int64(boma, Hash40::new("sword").hash as i64, Hash40::new("sword_thunder").hash as i64);
```
SkewedAskew: Duplicate of line 45?

fighters/robin/src/opff.rs
```diff
@@ -32,6 +32,26 @@ unsafe fn elwind1_burn(fighter: &mut L2CFighterCommon) {
...
+        fighter.on_flag(*FIGHTER_REFLET_INSTANCE_WORK_ID_FLAG_THUNDER_SWORD_ON);
+        VisibilityModule::set_int64(boma, Hash40::new("sword").hash as i64, Hash40::new("sword_thunder").hash as i64);
+        if WorkModule::get_int(boma, levin) == 1 {
```
SkewedAskew: This can be rewritten as follows:
```rust
if WorkModule::get_int(boma, levin) == 1 {
    app::FighterSpecializer_Reflet::set_flag_to_table(fighter.module_accessor as *mut app::FighterModuleAccessor, *FIGHTER_REFLET_MAGIC_KIND_SWORD, true, *FIGHTER_REFLET_INSTANCE_WORK_ID_INT_THROWAWAY_TABLE);
}
WorkModule::dec_int(boma, levin);
```

SkewedAskew (changes requested):

jaccob101:
Should have proofread before PRing x2 lmao ty for pointing everything wrong out

SkewedAskew (approved):

PR #1766, opened 2023-07-08
Controls screen crash fix

SkewedAskew:
Fixes an issue where the controls screen would crash if changing buttons after entering and leaving the test stage


CrossSockDance:
Nothing explodes save for the parry input swap.

WuBoytH (approved):
Gotta blame Frank on that one for sure

PR #1781, opened 2023-07-09
Allow all specials to fastfall

jobrien97:
Gives all special moves the ability to fastfall, with vanilla's fastfall conditions being:
- Must be descending (y velocity < 0.0)
- Must be able to drift horizontally (`FIGHTER_KINETIC_ENERGY_ID_CONTROL` enabled)

Note that these conditions mean many special moves will still not be able to fastfall, as many do not allow your character to drift during them (e.g. Kirby Inhale, Samus Charge Shot charge, Falcon Kick, etc.). If in the future, any of these were given the ability to drift, they would then be allowed to fastfall with this change.

This is mostly to address many recovery moves having an awkward period of time before reaching special fall where you cannot fastfall, though it looks like you should.

### DIFF
--- a/fighters/common/src/opff/physics.rs
+++ b/fighters/common/src/opff/physics.rs
@@ -61,9 +61,22 @@ unsafe fn plat_cancels(fighter: &mut L2CFighterCommon) {
     }
 }
 
+// Allows all special moves to be able to fastfall
+// with fastfall conditions being:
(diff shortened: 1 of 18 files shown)


jobrien97:
Can't be applied globally, too many edge cases. Will need to be done per-character

PR #1786, opened 2023-07-10
Double traction hotfix again (fix wavedash lengths)

jobrien97:
Wavedashes were made unintentionally very slightly shorter with the last double traction hotfix.


SkewedAskew (approved):
Here comes another

PR #1790, opened 2023-07-11
Bayonetta: Neutral B Cancel Fix

jaccob101:
Replaces change_status_req with change_status_force. Not sure about the exact differences but this seems to be higher priority and keeps bullets from cancelling into air dodge.


jobrien97 (changes requested):
please just use FALL, not FALL_AERIAL

jobrien97 (approved):

PR #1792, opened 2023-07-12
Pit Balancing

jdipalermo27:
Pit is a little overtuned, here's some changes to put him more in line with the cast.
Fair being a multihit that is safe rising and combos should not kill as early as it does, reducing the kbg should help with that. In return back air was buffed in bkb to increase it's kill potential.
Upair nerfed on kbg for the same reason.
Arrows have a small rework, increasing the speed and control of the arrow to keep the creative, freeflow aspect of the move, but nerfing the damage and life of the arrow to reduce it's sniping potential from far away offstage and increase the percent for when arrow sends into tumble.

<==Pit==>
Forward Aerial
~[-] KBG: 150 --> 140
Back Aerial
~[+] BKB: 30 --> 35
Up Aerial
~[-] KBG: 126 --> 114
Parameters
Arrow
~[R] Speed: 3.75 --> 6
~[R] Control Angle: 8 --> 20
~[-] Power: 4.5 --> 2
~[-] Life: 120 --> 25


SuddyN:
superseded by https://github.com/HDR-Development/HewDraw-Remix/pull/1894

PR #1794, opened 2023-07-12
Falcon Kick/Wizard's Foot/Power Thrust fixes

jobrien97:
### Falcon Kick:
- Can now slide when landing late into the animation

### Wizard's Foot:
- Can now slide when landing late into the animation

### Power Thrust:
- Does not slide nearly as far when landing immediately after diving


blu-dev (approved):
lgtm

PR #1799, opened 2023-07-14
Allow most specials to fast fall + remove aerial attack fastfall cancel

jobrien97:
Gives most special moves in the game the ability to fastfall, given that they meet vanilla's fastfall conditions:
- Must be descending (y velocity < 0.0)
- Must be able to drift horizontally (`FIGHTER_KINETIC_ENERGY_ID_CONTROL` enabled)

Note that these conditions mean many special moves will still not be able to fastfall, as many do not allow your character to drift during them (e.g. Kirby Inhale, Samus Charge Shot charge, Falcon Kick, etc.). If in the future, any of these were given the ability to drift, they would then be allowed to fastfall with this change.

This is mostly to address many recovery moves having an awkward period of time before reaching special fall where you cannot fastfall, though it looks like you should.

Also gives the following moves aerial drift to retain previous fastfall behavior:
- Link/Yink/Tink arrow
- PK Fire/PK Freeze

### **Also removes the Smash 4/Ultimate feature which cancels your fastfall when using an aerial.**


SkewedAskew (approved):
Hmm sure

PR #1801, opened 2023-07-14
Shield release FAF bugfix

jobrien97:
Fixes an issue where you could attack earlier than the intended FAF of shield release (f14).


Riah321:
Out of curiosity, what frame could you attack before?

WuBoytH (approved):

PR #1807, opened 2023-07-17
Slow down pummels

jobrien97:
Increases the time in between pummels, varying per character, but with a cap of 20f in between.

Standardizes all pummels to 2% damage.

Decreases **base grab hold time** from 90f -> 76f, and increases the **multiplier for grab hold time added from total damage** from 1.7 -> 2.2. (P+ values)


Riah321:
Why?

SkewedAskew (approved):

PR #1822, opened 2023-07-22
Pichu balance changes

stingers135:
Straightforward changes to bring Pichu in line with the rest of the cast.

Neutral Air
- Reduced hitbox size from 4.5 to 3.5 units
- Added 4 additional frames of endlag 
    - Autocancel window not affected

image

Down Air
- Reduced KBG from 110 (!) to 90

Forward Air
- KBG 100/85 -> 90/80

Dash Attack
- Changed angle to 361 (Sakurai angle)
- KBG 100/80 -> 85/65
     - With new angle, was killing too early.
- Reduced slide after hitboxes end

Forward Smash
 - KBG 110 -> 120

Up Smash
- KBG 105 -> 90
- Charge damage 18 -> 17

Thunder
- Changed weak hit angle from 94 to 115, to prevent situations where the weak hit combos too reliably into the strong hit
- Added meter gain to thunder when it connects with Pichu


hdr-balance-backroom (dismissed):
- Decrease forward air kbg slightly
- Decrease up smash kbg
- Revert dash attack endlag, apply `sv_kinetic_energy` to decrease the distance Pichu slides after the hitboxes clear
- Thunder should gain meter when it hits Pichu (add a set amount similar to neutral b)

stingers135:
> * Decrease forward air kbg slightly
> * Decrease up smash kbg
> * Revert dash attack endlag, apply `sv_kinetic_energy` to decrease the distance Pichu slides after the hitboxes clear
> * Thunder should gain meter when it hits Pichu (add a set amount similar to neutral b)

Changes made

H3ra-57 (approved):

PR #1831, opened 2023-07-24
Reintroduce heavy landing screenshake on non-wavedashes

jobrien97:
The heavies will cause a small screenshake on heavy landing again, but still not when wavedashing/wavelanding.


SkewedAskew (approved):
Yes

PR #1832, opened 2023-07-24
Shield release visual improvement

jobrien97:
Shield release now uses the whole animation instead of a cancel frame at motion frame 8.0, which then had to be motion rated to achieve an ingame FAF of 15, which caused shield releases to look veeeerryyy slloooooooww.


WuBoytH (approved):
I... don't think shield release looked particularly slow (even considering vanilla which has even less shield drop than HDR iirc)
however
functionally this looks fine

PR #1844, opened 2023-07-26
Fix spike crashes

jobrien97:
hopefully


Riah321:
Tested, LGTM

blu-dev (approved):
the structs here that are not used as interop between us and smash don't need to be repr c but idrc

PR #1851, opened 2023-07-28
C-stick fixes

### LINKED ISSUE #768: Cstick has a lockout window where you can't re-input shortly after releasing the stick
This prevents mashing the cstick to get an option out as soon as possible, which is a much larger issue in this game than vanilla since the buffer is so much weaker.

This is a thing for all cstick types except smash stick, which has no lockout time, and the lockout is external from the training mode frame advance (e.g. you won't notice it if you try to test in frame advance).

I believe the lockout was 8 frames in brawl of neutral cstick before you could re-flick the stick and get a result, but that may be wrong and I have no way to test the value in ult since it doesn't show with frame advance

### LINKED ISSUE #1027: Cstick down fastfalls your aerials
Inputting cstick down in the air has the same effect as fully slamming your main stick down and pressing A. It seems the cstick press makes you fastfall. An example:

Falcon is on the 15th airborne frame of his short hop.

manual slight down + A lands on frame 31
cstick down lands on frame 23
manual full down + a lands on frame 23 too

I also tested Marth and Incineroar and found the same pattern.

It is very noticable for Marth since it is impossible get a dair out with the cstick after an instant shorthop fair.

### LINKED ISSUE #1341: C-Stick Aerials check left stick for directional input under certain conditions.
When the C-stick and jump are input on the same frame, the game will only use the A input from the C-stick. When the aerial comes out, it will check which direction the left stick is being held and use that to determine which aerial to perform instead.

Example:
      1. Fair is input using jump and C-stick forward on the same frame (in the air or buffered during jumpsquat).
      2. Player wants to drift back, so they hold away on the left stick. When the player leaves the ground, Bair comes out instead of Fair 
      regardless of if C-stick forward is still held.

This has been in the game since Cstick drift was removed and it never got fixed (or wholly identified). This is probably, unironically, the leading cause of aerial misinputs in HDR.

### LINKED ISSUE #1703: 1 frame lockout for fastfall after any c-stick aerial
After performing an aerial attack _in any direction_ with the c-stick there is 1 frame where fast fall inputs won't be registered by the game

jobrien97:
- Removed the 10f lockout when attempting multiple inputs with Tilt Stick or Special Stick
- Buffered C-stick aerials no longer use the left stick's input in certain conditions
- Fixed downward C-stick inputs causing fastfall with Smash stick
- Removed 1f fastfall lockout after inputting a C-stick aerial
- Inputting C-stick + double jump on the same frame should no longer cause aerial drift

Fixes #768 
Fixes #1027 
Fixes #1341 
Fixes #1703


Riah321:
IT WORKS YEEEEAAAHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHH

PR #1858, opened 2023-07-31
Steve - Crafting Table Rework

moklmaru:
With the recent parry changes, Steve's ability to respawn his crafting table was removed to make way for the new parry input. This PR aims to restore that functionality by enabling Steve to respawn the table by pressing shield during the first 5 frames of standing mine. This is on parity with how Inkling refills their ink tank, essentially reversing the previous input from shield>special, to special>shield.

However, there are additional re-workings to how the table's behavior functions, as well as restrictions on when Steve can and can't summon it, per the request of the backroom. The following changes have been made to the mechanic:
- Steve can no longer summon the table if one already exists on the stage. Instead, the table has to be broken.
- When the table is broken, it no longer automatically respawns after a short amount of time. It will only reappear when manually summoned by Steve
- After a short duration of time following the table being broken, Steve will flash to indicate he can respawn it. 

These changes reintroduce Steve's ability to summon the table when needed, but reduces the amount of freedom he previously had to respawn it at will without any kind of commitment. Instead, choosing the location to put it will be more tactical and Steve will have to focus on breaking the table himself to regain the ability to do so, giving purpose to a previously redundant functionality of the vanilla mechanic.

Below is a video exampling Steve respawning the table with the new input (without restriction, no longer accurate to the PR).

https://github.com/HDR-Development/HewDraw-Remix/assets/16338870/ee368d7c-ffb6-4323-9966-149269979d7b

And here is a video roughly showcasing the new respawn functionality as it is in-game:

https://github.com/HDR-Development/HewDraw-Remix/assets/16338870/b835279b-95b4-49ea-ad86-c96dcf3028ac


hdr-balance-backroom:
This was an intentional nerf, we do not want Steve having the ability to move the table to a more advantageous position on a whim

hdr-balance-backroom:
After further discussion, we've decided on a different implementation in place of this mechanic

Steve should be able to place the crafting table but be unable to reposition it until it is destroyed. The table should not respawn randomly; instead, Steve should be able to create another table at his position after a brief period once it is destroyed (indicated by a gimmick flash once the cooldown is over)

The reason this was initially removed was because there was no importance placed on the position of the crafting table when Steve was able to reposition it with no commitment. With this design, Steve has more agency in where the table should be placed but will also need to think more about protecting it

SkewedAskew (approved):

PR #1865, opened 2023-08-02
Ryu Washes the Butter Off His Gloves

Riah321:
Set the SDI multiplier at the beginning of EX shoryuken to 0 so it doesn't drop 70% of the time.
Added a weakened late hit to air EX Shoryuken so that you can no longer perform "The Smash 4". (killing at 60% off the top)

Changed raw EX donkey kick to send out as a powerful launcher. If cancelled into, the combo continuation aspect has been improved instead.


SuddyN:
these changes will be integrated or replicated in https://github.com/HDR-Development/HewDraw-Remix/pull/1893

PR #1867, opened 2023-08-03
Nerf Wolf Aerial Shine Active Frames

jdipalermo27:
This must have been missed when making wolf shine frame 1, this changes aerial shine from 7 active frames to 1 active frame.


blu-dev (approved):
LGTM thanks a big one skadoooosh

PR #1870, opened 2023-08-05
Dedede fair rework

robertyur:
Reworked fair to be like brawl/pm

Current fair is very threatening due to its first active hitbox being above Dedede somewhat. This change aims to keep fair a strong move, but decrease its strength in situations where an opponent is above Dedede.

Forward Air:
Hitboxes reworked. Active on frames 14-20. Move ends on frame 39.
Hitbox 1: Damage 15%, Angle 361, KBG 95 FKB 0, BKB 15
Hitbox 2: Damage 13% Angle 361, KBG 95, FKB 0, BKB 15

[Video](https://streamable.com/6e8r16)

animation and config assets: 
[dededeassets.zip](https://github.com/HDR-Development/HewDraw-Remix/files/12447179/dededeassets.zip)


fighters/dedede/src/acmd/aerials.rs
```diff
@@ -56,39 +56,66 @@ unsafe fn dedede_attack_air_f_game(fighter: &mut L2CAgentBase) {
     let lua_state = fighter.lua_state_agent;
     let boma = fighter.boma();
     frame(lua_state, 1.0);
```
WuBoytH: Little bit of a nitpick but you can just get rid of the frame 1 stuff, no need to set the motion rate to 1.0 on frame 1

WuBoytH (dismissed):

fighters/dedede/src/acmd/aerials.rs
robertyur: Done!

WuBoytH: based

hdr-balance-backroom (dismissed):
- Motion rate the animation so the first active frame starts slightly later into the swing (it should not be hitting behind Dedede). Our suggestion is to use partial motion rates to get the start about halfway between the current first and second active frames

robertyur:
ba
Before -> After

SkewedAskew:
Can you move it forward a bit more? The leftmost part of the hitbox shouldn't be behind his center more than a unit or so

robertyur:
dedede
this better?

SkewedAskew:
Indeed, thanks

SkewedAskew (approved):

PR #1871, opened 2023-08-05
Ledge occupancy buff

jobrien97:
Ledge now counts as "occupied" for 90% of your ledge getup/attack/roll/jump animation, rather than 75%.


SuddyN (approved):
one line yea

PR #1872, opened 2023-08-07
Lucas Magnet Hitlag Parity 

jaccob101:
Matches Magnet hitlag to p+ for input parity. Now should be 1 hitlag on the hold hitbox, and 4 frames on the release hitbox. Previously the release had the 2 frames of hitlag and the hold hitbox had 5 frames. Counting the frame it hits.


jaccob101:
made redundant

PR #1873, opened 2023-08-07
Fix Terry/Ken/Ryu/Yoshi shield release FAF

jobrien97:
Should now be f15 like the rest of the cast.


SuddyN (approved):
erm I actually oppose this change for reasons

PR #1875, opened 2023-08-08
Slightly slow down ledge attacks

jobrien97:
Applies a 0.92 motion rate to all ledge attacks, which pushes back their hitbox timing/FAF to be more in line with P+.


SkewedAskew (approved):

SuddyN:
It works, but do we really want getup attack to be even more unsafe?

jobrien97:
> It works, but do we really want getup attack to be even more unsafe?

This is for ledge attack, not getup attack

SuddyN:
> This is for ledge attack, not getup attack

thats what I meant

PR #1876, opened 2023-08-08
Fix shield release anims again lol

jobrien97:
Speed of shield release anims should be more consistent across the cast again.


SkewedAskew (approved):
:twelve:

SuddyN:
looks good in game

PR #1884, opened 2023-08-09
More sliding/disjointed move bugfixes

jobrien97:
Reverts the `move` flag to false for moves which use hitboxes on the `top` bone.


SkewedAskew (approved):
false

PR #1887, opened 2023-08-10
Increase minimum camera zoom

jobrien97:
The camera is further zoomed out by default now.
Global min camera zoom value increased 120 -> 145

Before:

2023081013143400-0E7DF678130F4F0FA2C88AE72B47AFDF

After:

2023081013092800-0E7DF678130F4F0FA2C88AE72B47AFDF

To be merged with HDR-Development/hdr-private#435


SuddyN (approved):
needs more comments

PR #1904, opened 2023-08-20
Resolve dependency error

SkewedAskew:
Fixes (hopefully) a dependency issue resulting from a skyline-web update that was causing builds to break


WuBoytH (approved):
I mean it built, so...

PR #1907, opened 2023-08-20
Restores Ryu/Ken's Fixed Jump Arcs

WuBoytH:
In vanilla, Ryu and Ken kept their horizontal jump speed if they release the control stick. This was absent due to an oversight in the control energy update function, and should be restored.

Tested in WuBor, where the same issue was occurring.


SuddyN (approved):
makes sense and seems to work

PR #1927, opened 2023-08-30
Palutena bugfix

H3ra-57:
Fixes issue where the wind sound effect from Spin Attack would continue playing if interrupted


SkewedAskew (approved):
Does thing

PR #1929, opened 2023-08-31
Snake buffs

jdipalermo27:
After the nerfs and PM move additions to snake, snake is feeling a bit more inconsistent and weaker, this pr aims to fix these inconsistencies and push back some of the nerfs to his aerials in the other direction.

Forward smash being a 3 hit move produces a lot of issues in consistency, in order to keep the 3 hits and solve this issue, I believe making all 3 hits launchers would keep the move fair, consistent, useful, and unique to multihit smash attacks.

Nair hitboxes are inconsistent in how they connect since the angle changes, so I tweaked the knockback on the moves in order to help them connect more consistently. I started with trying to match the connection hitboxes with zss' first hit of fair, but I will be tweaking the values as I continue to test.

When I nerfed bair, I both nerfed the activity and the kb, the activity feels very good now, but the kb is lackluster. I increased the damage on the inner hitboxes to give bair that log feeling I've viewed it to have and have consistent kb across the whole move.

Tranq needed very quick nerfs after being added to the game, now no longer sleeping opponents in the air. The struggle of this is that it has no use in the air now, so I changed the collision type to apply hitstun in the air (this fix may not be right, i may need to increase the range or something else)

<==Snake==>
FORWARD SMASH
Hitbox 1
Damage: 6.0 --> 7.5
KBG: 25 --> 75
BKB: 20 --> 85
Hitbox 2
Damage: 6.0 --> 7.5
KBG: 25 --> 75
BKB: 20 --> 85
NAIR
Multihits
~[R] KBG: 100 --> 180
~[R] FKB: 50 --> 35
BAIR
Strong Hitbox ID 0
~[+] Damage: 12 --> 14
Strong Hitbox ID 1
~[+] Damage: 13 --> 14
Tranq
Hitbox ID 2
~[+] Collision Type: collision_attr_sting --> collision_attr_normal


jdipalermo27:
I'm going to switch back dart, the rest should be good for now

PR #1930, opened 2023-08-31
Slow pummel near-reversion

jobrien97:
Pummel speed has been reverted to basically current beta behavior.

They have instead been given a high global hitlag multiplier to better emulate vanilla's pummel impact.

### DIFF
--- a/fighters/bayonetta/src/acmd/other.rs
+++ b/fighters/bayonetta/src/acmd/other.rs
@@ -660,7 +660,7 @@ unsafe fn game_catchattack(agent: &mut L2CAgentBase) {
     }
     frame(lua_state, 1.0);
     if is_excute(agent) {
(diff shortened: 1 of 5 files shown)


SuddyN (commented):
It looks like this still uses the frame cap on pummel animations you previously introduced. Will that result in some pummels that were previously slow (e.g. Kazuya) to be faster in practice?

jobrien97:
> It looks like this still uses the frame cap on pummel animations you previously introduced. Will that result in some pummels that were previously slow (e.g. Kazuya) to be faster in practice?

No, setting the frame cap to 8 essentially makes the FAF 8 (which was the global pummel FAF in vanilla). Rather than editing every single motion patch again, I just did it this way

SkewedAskew (approved):

PR #1938, opened 2023-09-02
Ganondorf P+ Dthrow

Odd-Rick:
To make Ganondorf's dthrow more useful than just a downb follow up, I've changed the values of dthrow to match up with Project+ Ganondorf dthrow. 

Edit:
Adjusted the angle from 80 to 78, the angle used in P+, as to not make it as easy to chain grab, due to the amount of hitstun in HDR's engine.

**Changelog:**

**Down Throw:**

- [+] Angle: 78 (Changed from 64)
- [/] BKB: 45 (Changed from 50)
- [/] KBG: 75 (Changed from 70)

### DIFF
--- a/fighters/ganon/src/acmd/specials.rs
+++ b/fighters/ganon/src/acmd/specials.rs
@@ -122,6 +122,36 @@ unsafe fn ganon_float_eff(fighter: &mut L2CAgentBase) {
     }
 }
 
+#[acmd_script( agent = "ganon", script = "game_specialairsstart", category = ACMD_GAME, low_priority )]
+unsafe fn game_specialairsstart(agent: &mut L2CAgentBase) {
+    let lua_state = agent.lua_state_agent;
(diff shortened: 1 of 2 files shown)


yoonikoo:
What are these values even changed from? What were the original values??

Odd-Rick:
> What are these values even changed from? What were the original values??

The original angle was 65, the BKB was originally 50, and the KGB was originally 70.

yoonikoo:
> > What are these values even changed from? What were the original values??
> 
> The original angle was 65, the BKB was originally 50, and the KGB was originally 70.

List these in the changelog

Odd-Rick:
> > > What are these values even changed from? What were the original values??
> > 
> > 
> > The original angle was 65, the BKB was originally 50, and the KGB was originally 70.
> 
> List these in the changelog

Alright, done.

PR #1979, opened 2023-09-27
Rivals Wall Jump followup

SkewedAskew:
Shortens the input window for a valid wall jump to 3 frames, making button wall jumps more difficult to hold buffer. This should mitigate accidentally performing a wall jump when performing actions such as rising aerials from ledge

### DIFF
--- a/utils/src/modules/input.rs
+++ b/utils/src/modules/input.rs
@@ -549,18 +549,12 @@ fn exec_internal(input_module: &mut InputModule, control_module: u64, call_origi
         && triggered_buttons.intersects(Buttons::Jump) {
             if ControlModule::get_stick_x((*input_module.owner).module_accessor) > 0.0 {
                 if input_module.hdr_cat.valid_frames[walljump_right_offset] == 0 {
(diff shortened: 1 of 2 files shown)


utils/src/modules/input.rs
```diff
@@ -549,18 +549,12 @@ fn exec_internal(input_module: &mut InputModule, control_module: u64, call_origi
...
             }
             else if ControlModule::get_stick_x((*input_module.owner).module_accessor) < 0.0 {
                 if input_module.hdr_cat.valid_frames[walljump_left_offset] == 0 {
-                    input_module.hdr_cat.valid_frames[walljump_left_offset] = unsafe {
-                        ControlModule::get_command_life_count_max((*input_module.owner).module_accessor)
-                            as u8
-                    };
+                    input_module.hdr_cat.valid_frames[walljump_left_offset] = 2;
                 }
             }
```
SuddyN: parameterize this?

SkewedAskew: I will once this is tested and verified

WuBoytH (approved):

PR #2002, opened 2023-10-06
Small Duck Hunt buffs

jdipalermo27:
Unfinished and in testing.

Some changes to even him out a little bit around the current direction he's being pushed.

Bkb increase and kbg decrease so the move is safer on hit without increasing kill power.
Decrease frames between hits of fsmash to match the pace of the game without making the move come out faster.
Decrease in faf of side b as the move is almost useful atm. Right now it's huge risk with almost no reward, at best it leads to a tech chase. This will allow it to be a strong combo starter and setplay tool, but it keeps the higher startup to make it easy to punish when whiffed in close to medium ranges.

<==Duck Hunt==>
NAIR
Hitbox
~[+] BKB: 30 --> 40
~[-] KBG: 95 --> 85
UPSMASH

FSMASH

DSMASH

CLAY PIGEON
Grounded
~[+] FAF: 68
Aerial
~[+] FAF: 68


WuBoytH:
Not being considered, adjustments are planned for a Misc. Updates PR.

PR #2020, opened 2023-10-15
Fixes UI Crashing in Mii Maker

WuBoytH:
Does what it says on the tin.


blu-dev (approved):
LGTM, so long as it works (which it defo does)

PR #2023, opened 2023-10-16
Pikachu - single hit nair

jobrien97:
Stats taken from P+:
- Damage (early/late) 12%/9%
- Angle 361
- KBG 100
- BKB (early/late) 18/0
- AC 0-3/35-40
- LL 7f


SkewedAskew:
Revised and added to misc adjustments

PR #2032, opened 2023-10-20
Remove 5f tap buffer extension during tumble

jobrien97:
Removes the training wheels

Tap buffer during tumble 10f -> 5f

### DIFF
--- a/fighters/common/src/general_statuses/mod.rs
+++ b/fighters/common/src/general_statuses/mod.rs
@@ -90,7 +90,7 @@ pub unsafe fn status_pre_DamageAir(fighter: &mut L2CFighterCommon) -> L2CValue {
 
 #[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_sub_DamageFlyCommon_init)]
 pub unsafe fn damage_fly_common_init(fighter: &mut L2CFighterCommon) {
-    ControlModule::set_command_life_extend(fighter.module_accessor, 0);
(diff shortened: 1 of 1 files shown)


fighters/common/src/general_statuses/mod.rs
```diff
@@ -90,7 +90,7 @@ pub unsafe fn status_pre_DamageAir(fighter: &mut L2CFighterCommon) -> L2CValue {
 
 #[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_sub_DamageFlyCommon_init)]
 pub unsafe fn damage_fly_common_init(fighter: &mut L2CFighterCommon) {
-    ControlModule::set_command_life_extend(fighter.module_accessor, 5);
+    ControlModule::set_command_life_extend(fighter.module_accessor, 0);
```
WuBoytH: this line shouldn't even exist if you wanted to get rid of this

WuBoytH (changes requested):

fighters/common/src/general_statuses/mod.rs
jobrien97: u get a comment out, take it or leave it

WuBoytH (approved):
bringus

fighters/common/src/general_statuses/mod.rs
O1yyy: lgtm

PR #2040, opened 2023-10-27
B-Input Gentlemans for Rapid Jabbers

jaccob101:
Characters with rapid jabs and multi-stage jabs can press B to gentleman. This matches the behavior added to single-hit rapid jabbers such as Zelda.


WuBoytH:
Currently not being considered, and would rather implement them as jab 3s instead of pressing the Special button.

PR #2044, opened 2023-10-30
Piranha Plant changes pt. 3.5

H3ra-57:
### Adaptive Roots
- (!) Instead of taunts manually choosing a stance, taunting now cycles through stances in the order Standard → Putrid → Prickly. If you are holding a taunt button on frame 8, you go backwards in the cycle.
- [-] Stance Change frame: F1 → F8
- [+] Up taunt now has 10% heavy armor F1-F9


SuddyN (dismissed):
Can you implement a button hold check f1-8 instead of settling for only checking it on one frame? 

I'm going to be re-implementing hold checks on Lucario supers and (upcoming) PT switching to check hold inputs like Ken/Ryu short/long press normals. Players have repeatedly expressed frustrations about hold-versions triggering accidentally on double presses.

H3ra-57:
> Can you implement a button hold check f1-8 instead of settling for only checking it on one frame?
> 
> I'm going to be re-implementing hold checks on Lucario supers and (upcoming) PT switching to check hold inputs like Ken/Ryu short/long press normals. Players have repeatedly expressed frustrations about hold-versions triggering accidentally on double presses.

it has been done

fighters/packun/src/acmd/other.rs
```diff
@@ -128,21 +128,87 @@ unsafe fn appeal_hi_game(fighter: &mut L2CAgentBase) {
     let lua_state = fighter.lua_state_agent;
     let boma = fighter.boma();
     let cur_stance = VarModule::get_int(boma.object(), vars::packun::instance::CURRENT_STANCE);
-    VarModule::set_int(boma.object(), vars::packun::instance::CURRENT_STANCE, (cur_stance + 1) % 3);
-    VarModule::on_flag(fighter.object(), vars::packun::instance::STANCE_INIT);
-    frame(lua_state, 10.0);
+    let mut advance = 0;
+    let mut holding = true;
```
SuddyN: I believe you'll need to use VarModule to properly track the `holding` flag e.g. `vars::packun::status::is_button_held`

fighters/packun/src/acmd/other.rs
```diff
@@ -128,21 +128,87 @@ unsafe fn appeal_hi_game(fighter: &mut L2CAgentBase) {
...
+        }
+    }
+    frame(lua_state, 7.0);
+    if is_excute(fighter) {
+        if holding {
+            if !boma.is_button_on(Buttons::AppealAll) {
+                holding = false;
+            }
+        }
+    }
```
SuddyN: You can use a loop like this to avoid repeating code

```rs
frame(lua_state, 1.0);
if is_excute(fighter) {
  // set the flag to true
  // anything else you need to do frame 1
}
for _ in 1..7 {
  if is_excute(fighter) {
    // set flag false if not holding
  }
  wait(lua_state, 1.0); // wait happens last, meaning the above block also runs on frame 1
}
frame(lua_state, 8.0);
if is_excute(fighter) {
  // check the hold-flag and choose a stance
}
```

fighters/packun/src/acmd/other.rs
```diff
@@ -128,21 +128,87 @@ unsafe fn appeal_hi_game(fighter: &mut L2CAgentBase) {
...
+    frame(lua_state, 8.0);
+    if is_excute(fighter) {
+        if holding {
+            advance = 2;
+        }
+        else {
+            advance = 1;
+        }
```
SuddyN: since you only use the `advance` var here, please define it here too (instead of at the top of the function)

generally, you should only define variables as early as you need to reference them

SuddyN:
```rs
let advance = if holding {2} else {1};
```

fighters/packun/src/acmd/other.rs
```diff
@@ -156,31 +222,79 @@ unsafe fn appeal_hi_sound(fighter: &mut L2CAgentBase) {
     }
 }
 
-#[acmd_script( agent = "packun", script = "game_appealsl", category = ACMD_GAME , low_priority)]
-unsafe fn appeal_s_l_game(fighter: &mut L2CAgentBase) {
+#[acmd_script( agent = "packun", scripts = ["game_appealsl", "game_appealsr"] , category = ACMD_GAME , low_priority)]
```
SuddyN, line 166: repeat my review for the other taunts

SuddyN (dismissed):

fighters/packun/src/acmd/other.rs
WuBoytH: this is so ugly I genuinely would rather you handle it via OPFF

SuddyN: I think normally this would be done in status code and the resulting plant transformation form would just be passed in via workmodule

WuBoytH: I just don't think ACMD should be used like this, it doesn't feel like good practice
do an OPFF, do a taunt status and motion check, set a flag to true on frame 1 of ACMD and if the button is off before frame X turn off the flag.

PR #2048, opened 2023-11-01
Tap buffer reduction

TeamPearson:
Tap buffer size 5F>3F


WuBoytH:
Currently not being considered

PR #2049, opened 2023-11-01
Remove Gatlings

SuddyN:
DACUS/DACDS can no longer be performed after dash attack connects on hit or shield


SkewedAskew (approved):
Perish

PR #2051, opened 2023-11-02
Recovery adjustments part X ((WIP))

jobrien97:
Aims to reduce the horizontal potency of many recovery moves, to compensate for the horizontal recovery ability of HDR's airdodges.

Full changelog coming soon


SkewedAskew:
Will be reevaluated after upcoming changes

PR #2053, opened 2023-11-02
Per player buffer

blu-dev:
UNDECIDED

`Low` is 1f of buffer
`Normal` is 5f of buffer (default)
`High` is 3f of buffer

The UI would need to be updated but the functionality is there


SkewedAskew:
Not being considered in favor of refining the current buffer

PR #2060, opened 2023-11-09
Young Link; Downward Edition

Riah321:
Some changes to bring back his more inspired downward move attributes from melee. With our scientific research we have finally managed to create a zoner with sauce! I'm sure some numbers will need tweaking, but yes dair IS supposed to be that crazy.

Changes to:
Down Tilt: Body and hilt hitboxes now spike ala melee
Down Smash: Innermost hitbox sends up.
Down Air: The late hit is now the strong hit. It's also hella juiced like in melee.
Down Throw: Pops up very low for techchase on spacies and combos on normal characters.


SkewedAskew:
Not being considered

PR #2078, opened 2023-11-22
Dark Samus recovery adjustment

jobrien97:
### UpB:

- Rising max horizontal air speed reduced 0.57 -> 0.2


SkewedAskew:
Will be reevaluated after upcoming changes

PR #2080, opened 2023-11-23
Add sound effect to stage strike

sticks-stuff:


SkewedAskew:
Would you please provide a video of this?

PR #2096, opened 2023-11-28
Global FAF ledgegrab

jobrien97:
Allows all special moves to grab ledge **once their FAF is reached**, rather than **when their animation completes**. The difference this makes will largely vary between each character and move, as it depends on how many frames exist between the move's FAF and the end of its animation.


fighters/common/src/opff/ledges.rs
```diff
@@ -100,8 +101,29 @@ unsafe fn tether_trump_landing(boma: &mut BattleObjectModuleAccessor, status_kin
...
+unsafe fn global_faf_ledgegrab(fighter: &mut L2CFighterCommon) {
+    // If a special move has a defined FAF (!= 0)
+    // then we allow it to grab ledge
+    if StatusModule::status_kind(fighter.module_accessor) > 0x1DB  // only applies to special moves
+    && fighter.is_situation(*SITUATION_KIND_AIR)
+    && {let motion_kind = MotionModule::motion_kind(fighter.module_accessor);
+        FighterMotionModuleImpl::get_cancel_frame(fighter.module_accessor, Hash40::new_raw(motion_kind), true) > 0.0}
+    {
+        // Both needed to be able to grab ledge
+        if fighter.status_frame() == 1 {
```
WuBoytH: this implementation will cause gameplay changes for moves that already have `sub_transition_group_check_air_cliff` in its code, allowing them to grab ledge at any point while falling. This should be reimplemented as the following to keep the move's normal ledge grab behavior:
```rs
if cancel_frame > 0.0 {
    if MotionModule::frame == cancel_frame {
        fighter.sub_fighter_cliff_check(GROUND_CLIFF_CHECK_KIND_ON_DROP_BOTH_SIDES.into());
    }
    if is_enable_cancel... {
        sub_transition_group_check_air_cliff
    }
}
```

WuBoytH (dismissed):

WuBoytH (changes requested):
see previous comment

fighters/common/src/opff/ledges.rs
jobrien97: Done, but did not use an == check to prevent any motion rated moves from failing this check

WuBoytH (approved):
Alright this should be OK
before it gets merged someone should test if it makes certain moves (Ryu Hadoken at ledge for example) more toxic to deal with

PR #2107, opened 2023-12-04
Kirby NeutralB Bugfixes

ethangarsia517:
Provides several fixes to some Kirby's many abilities.

- Fixes an issue where Kirby, in the air, could not shield cancel chargeable specials if he had previously used his airdodge in the same airtime. This is the same effect as this PR, but for Kirby's copy abilities.
https://github.com/HDR-Development/HewDraw-Remix/pull/2092

- Fixes an issue with Dr. Mario's copy ability where, the normal Mega Vitamins animation would not play correctly if you had previously lost the copy ability, and then copied it again in the same match or training session.

- Fixes an issue where Kirby was able to B-Reverse, Reverse-B, Wavebounce, or otherwise change momentum/direction twice with Inhale in a single use.

- Allows Kirby to B-Reverse or Wavebounce the charging animation of Little Mac's copy ability Straight Lunge.


SkewedAskew (commented):
Lgtm

PR #2143, opened 2023-12-20
Fix Richter crouching ftilt

jobrien97:
Fixes Richter's crouching ftilt being visually bugged and missing hitboxes.


SkewedAskew (approved):
Consider adding the ability to transition going into crouch? Otherwise lgtm

PR #2152, opened 2023-12-31
Script to rename cracked hashes

zrksyd:
Adds a script that will go through ParamLabels.csv (must specify path) and finds all hashes and will replace already cracked hashes with their string counterpart. It will print all cracked and uncracked hashes afterwards. Must require some manual editing but it works for the most part.


WuBoytH:
please remake this PR, i should not be seeing gameplay changes in the diffs

PR #2170, opened 2024-01-13
DJC Leniency

### LINKED ISSUE #1888: Various DJC Leniency Window bugs
with djc characters, you can hold Y after double jumping during your aerial to get a rising aerial using your dj movement arc, but it seems like i've nailed a few edge cases for the behavior of this mechanic
- only really tested with lucas!

## momentum flings

https://github.com/HDR-Development/HewDraw-Remix/assets/19886440/4c2fea67-cb98-4b36-8e25-463d0d7aaaf1

this happens if you DJ and hold Y, let it run for a few frames (i went through 8 frames of the dj here, inputted fair on f5) hold Y for the duration of hitlag, then let go once the hitlag is done

the weird part is that djc leniency frame param is set to 6, meaning the expected behaviour is allow the dj's momentum to carry over for only the first 6 frames if you're holding Y during your aerial still, and if you go past that you will have to run through the entire arc, giving you a rising dj aerial

and here i'm hitting fair's hitbox on f9 of the double jump, WAY past the f6 param, therefore i should expect to go through the whole dj arc, but it allows me to still inheret the latter fast portion of the dj, giving that behaviour

my guess culprit is this piece here:
Code_2023-08-10-23-35-33

or
Code_2023-08-10-23-30-53

both of these run the same check, wherein i suspect `CURRENT_FRAME` starts counting from 0 again when you enter another status, in this case forward air

ig to fix this you could set a "last maximum frame" count that just keeps track of the highest amount of frames you went through the DJ with, then adding that last known value to the current status's frame, should disallow the behavior altogether
if theres a way to access a "how many frames the last status ran for" count, that could probably do it

## leniency window momentum inconsistencies

https://github.com/HDR-Development/HewDraw-Remix/assets/19886440/38d888b0-1371-405f-b0f6-5854b8ceb7bf

f1 - Y (jump)
f2
f3
f4 - Y (buffer jump in jumpsquat)
f5 - Hold Y {{lucas enters dj}}
f6 - Hold Y, *input uair or fair* 
f7 - **let go of Y**
f8 - **or let go of Y here**

results in landing sooner than expeceted, its somehow lessening the previous Y momentum
or rather it doesnt look like it carries it over at all

my guess is the leniency code, notice im holding Y while also inputting my aerial, meaning that whenever i let go of Y i no longer inheret the dj arc
normally you should have left over speed from coming out of the dj arc, but i actually think this is happening because we cancel out of uair's animation, not the dj itself?
like the inertia from dj cancels out all together after you let go of Y **if you have already entered an aerial while still holding Y**
the moment you let go of Y during an aerial, its like that momentum is almost 0'd out, not carried over, or at least not the same way as simply canceling out of the dj animation

- tweaking the dj arc so it gains more momentum in the first few frames did NOT fix the issue
this looks like its exclusively the fault of the djc leniency window 
- **this does not happen if you simply input an aerial without Y held, simply inputting an aerial while in dj carries over the proper momentum,** but in the leniency window its seemingly inheriting less momentum

https://github.com/HDR-Development/HewDraw-Remix/assets/19886440/9c876d4f-5418-48bb-8905-49c37a949a45

in pm this isnt an issue, it still moves through the dj arc for one more frame (?) as it would if you were to just cancel dj with an aerial
note the diamond on the very bottom of lucas, thats the floor collision

so to fix this i believe the leniency window code should follow the arc for one more frame (???) or match the cancel behavior of inputting an aerial to cancel dj
0 idea, i just know that i tried what i could and i dont know how to fiddle with the djc.rs file at all...
i'm probably guessing this behaviour is due to fixing one frame physics delay

jaccob101:
Due to engine differences djcs out of shorthops gain far less height, and inputting dj+aerial at the same time gains 0 jump momentum. In addition, there appears to be a bug where you can superjump during hitlag. This attempts to help with those issues, in lieu of an in-depth look by a dev. Tested with all djcers, dj arcs adjusted on ness/peach. Originally was going to be only code changes, however since I cannot properly fix the problem I additionally made the pre-existing jumps better (other than lucas)

Resolves #1888
Assets: [adjusted djc.zip](https://github.com/HDR-Development/HewDraw-Remix/files/14016611/adjusted.djc.zip)

### DJC:
- [*] Releasing jump during a rising aerial now waits for hitlag to end, preventing some odd behaviors around the leniency window mentioned in issue (from my testing)
- [/] Djcs now wait until 1 frame of jump momentum has passed, and can cancel on frame 2+ of the arc
- [*] DJC check now only in exec, before it was in main, main_loop, and exec at the same time?

**Peach:**
- [/] Dip portion of dj widened, so that you can get perfect djcs even with the above change. However the curve is less steep, so you don't get sucked down as fast without manually fast falling.

**Ness:**
- [/] DJ arcs from brawl/pm have been ported, except with the rise portion starting 4 frames sooner and frames 6/7 gaining slightly more momentum. This enables faster-paced djc combos while not burdening him with the full slowness of his melee dj

**Lucas:**
- [*] DJ anim from before Oly's pr has been restored, no longer pushed forwards a frame (did not solve any issues)

### DIFF
--- a/fighters/common/src/djc.rs
+++ b/fighters/common/src/djc.rs
@@ -23,13 +23,6 @@ pub unsafe extern "C" fn attack_air_main_status(fighter: &mut L2CFighterCommon) 
 /// Performs the leniency check for double jump canceling
 #[utils::export(common::djc)]
 pub unsafe extern "C" fn attack_air_main_status_loop(fighter: &mut L2CFighterCommon) -> L2CValue {
-    if fighter.global_table[PREV_STATUS_KIND] == FIGHTER_STATUS_KIND_JUMP_AERIAL {
-        let djc_leniency_frame = ParamModule::get_int(fighter.battle_object, ParamType::Common, "djc_leniency_frame");
-        if fighter.global_table[CURRENT_FRAME].get_i32() <= djc_leniency_frame && !ControlModule::check_button_on(fighter.module_accessor, *CONTROL_PAD_BUTTON_JUMP) {
-            WorkModule::on_flag(fighter.module_accessor, *FIGHTER_INSTANCE_WORK_ID_FLAG_JUMP_NO_LIMIT_ONCE);
-            KineticModule::change_kinetic(fighter.module_accessor, *FIGHTER_KINETIC_TYPE_MOTION_FALL);
-        }
-    }
     if !fighter.status_AttackAir_Main_common().get_bool() {
         fighter.sub_air_check_superleaf_fall_slowly();
         if !fighter.global_table[IS_STOPPING].get_bool() {
@@ -54,12 +47,9 @@ pub unsafe extern "C" fn sub_attack_air_inherit_jump_aerial_motion_uniq_process_
 
         fighter.sub_attack_air_kind();
         if motion_kind == smash::hash40("jump_aerial_f") || motion_kind == smash::hash40("jump_aerial_b") {
-            if !WorkModule::is_flag(fighter.module_accessor, *FIGHTER_INSTANCE_WORK_ID_FLAG_IGNORE_2ND_MOTION)
-            && ControlModule::check_button_on(fighter.module_accessor, *CONTROL_PAD_BUTTON_JUMP) {
+            if !WorkModule::is_flag(fighter.module_accessor, *FIGHTER_INSTANCE_WORK_ID_FLAG_IGNORE_2ND_MOTION) {
                 MotionModule::add_motion_2nd(fighter.module_accessor, Hash40::new_raw(motion_kind), frame, 1.0, false, 1.0);
                 MotionModule::set_weight(fighter.module_accessor, 1.0, true);
-            }
-            if ControlModule::check_button_on(fighter.module_accessor, *CONTROL_PAD_BUTTON_JUMP) {
                 KineticModule::change_kinetic(fighter.module_accessor, *FIGHTER_KINETIC_TYPE_JUMP_AERIAL_MOTION_2ND);
             } else {
                 WorkModule::on_flag(fighter.module_accessor, *FIGHTER_INSTANCE_WORK_ID_FLAG_JUMP_NO_LIMIT_ONCE);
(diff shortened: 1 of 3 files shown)


jobrien97:
You're gonna wanna use the `is_in_hitlag` boma ext rather than `StopModule::is_stop` here. `is_stop` only returns true when an attack receiver is in hitlag.

fighters/common/src/djc.rs
```diff
@@ -56,7 +59,7 @@ pub unsafe extern "C" fn sub_attack_air_inherit_jump_aerial_motion_uniq_process_
         if motion_kind == smash::hash40("jump_aerial_f") || motion_kind == smash::hash40("jump_aerial_b") {
             if !WorkModule::is_flag(fighter.module_accessor, *FIGHTER_INSTANCE_WORK_ID_FLAG_IGNORE_2ND_MOTION)
             && ControlModule::check_button_on(fighter.module_accessor, *CONTROL_PAD_BUTTON_JUMP) {
-                MotionModule::add_motion_2nd(fighter.module_accessor, Hash40::new_raw(motion_kind), frame, 1.0, false, 1.0);
+                MotionModule::add_motion_2nd(fighter.module_accessor, Hash40::new_raw(motion_kind), frame.clamp(1.0, MotionModule::end_frame(fighter.module_accessor)), 1.0, false, 1.0);
```
jobrien97: This could just be `frame.max(1.0)`.

jobrien97:
Also, similarly to Peach, you can find Sora's DJC code in fighters/trail/src/status/attack_air.rs

fighters/peach/src/status/attack_air.rs
```diff
@@ -92,6 +92,8 @@ unsafe extern "C" fn peach_attack_air_no_float_main_loop(fighter: &mut L2CFighte
 
     
     if fighter.global_table[CURRENT_FRAME].get_i32() <= ParamModule::get_int(fighter.battle_object, ParamType::Common, "djc_leniency_frame")
+    && fighter.global_table[CURRENT_FRAME].get_i32() > 1
+    && !StopModule::is_stop(fighter.module_accessor)
```
jobrien97: `!fighter.is_in_hitlag()` pls

jaccob101: im sorry i didn't pull before pushing and it did that

PR #2177, opened 2024-01-15
Jab sweep

jobrien97:
Aims to give almost every character a pop-up jab, in accordance with older games, getting rid of the 361 angle most jabs were given in Ultimate.

Several veterans' jab stats pulled from either P+ or Smash 4.

Also sets a global SDI multiplier of 0.3 to rapid jabs.

Full changelog coming soon


WuBoytH:
Not being considered at this time.

PR #2178, opened 2024-01-15
Incineroar - sideB freefall

jobrien97:
Alolan Whip now sends into freefall on whiff.

No longer once-per-airtime.


SkewedAskew:
Not being considered

PR #2184, opened 2024-01-16
Neutral getup input improvement

jobrien97:
A shield input when knocked down will now trigger neutral getup.


WuBoytH:
Unsure if this should make it into the game in the case of missing a tech and getting up earlier than intended. Should just be left to tilting the stick upwards imo

WuBoytH:

PR #2192, opened 2024-01-19
Ledge occupancy buff

jobrien97:
Ledge now counts as "occupied" until your ledge getup/attack/roll/jump FAF, rather than FAF * 0.9 frames.


WuBoytH:
Alright everybody, count how many times will JOB buff ledgehogging within one year

PR #2204, opened 2024-01-20
(even more) Copy Ability Adjustments

ethangarsia517:
An addition and an adjustment for two of Kirby's, copy abilities.

**Piranha Plant's Ptooie**
[!] Kirby is now able to use the Putrid and Prickly stances. He will always copy the stance that Piranha Plant is currently using when swallowed. He is unable to change stances by himself. The only way he can change stances is by losing the ability, and swallowing Piranha Plant while she is in a different stance.

**Diddy Kong's Peanut Popgun**
[-] Kirby's shield and jump canceling of Peanut Popgun charge is no longer frame 1, and is identical to Diddy Kong's.


SkewedAskew:
I would rather that Kirby just copy the properties of whatever stance Plant was in when inhaled, as being able to swap without first discarding is inconsistent with other form change characters

PR #2206, opened 2024-01-22
Shield Mechanics Rewrite

SuddyN:
## General
- The static shield effect which previously displayed during shield hitlag will now persist until actionable again
  - When remaining shield lag is 10 frames or less, the effect will start to become transparent with alpha = remaining-frames/10

## Completely rewrote out-of-shield logic
- can now hold-buffer spotdodge/roll OOS with the c-stick on tilt/smash/special/grab stick
- can now up special OOS without a flick input (like any other up special input)

options available when not shield locking, in order of priority
- buffered aerial
- up smash / upwards item toss
- up special
- jump
- spotdodge
- roll
- item toss (not with c-stick)
- plat-drop
- parry

options available when shield locking, in order of priority
- buffered aerial
- up smash / upwards item toss
- up special
- jump (not tap jump)
- item toss (including c-stick)
- plat-drop (with simple down input)
- parry


stingers135:
Holding the c stick does not continuously buffer rolls/spotdodges. It only works for the first one if doing it from held shield.

If buffering shield + c stick from another action, you get a grab instead of a roll/spotdodge.

PR #2207, opened 2024-01-22
Roy Uair Nerf

LunchablesTX:
The goal of these changes are to make up air slightly less over-centralizing. The hitbox priority on Up Air has been swapped to reduce the amount of sweetspot up airs that can be chained in a single combo. The hitbox priority now requires you to hit the inner two hitboxes (Body or Arm) to hit the sweetspot. The reduced knockback on the tipper hitbox still allows for combos, such as tipper Uair > Jab. This should hopefully keep Up Air as a strong combo tool that requires different combo routes rather than just looping itself.

I would like to do many other changes to this character, but even without the context of any new moves/etc I still think up air could use some adjustments.

Up Air:
~ [-] Landing Lag: 8 -> 9
~ [-] BKB (Tipper): 35 -> 30
~ [-] KBG (Tipper): 70 -> 60
~ [-] Damage (Tipper): 8 -> 7
~ [-] Angle (Tipper): 90 > 80
~ [-] Hitbox priority changed

- Previously:

Sword > Arm > Body > Tip

- Now:

Arm > Body > Tip > Sword

Screenshot_2
image


elias-michaias:
The code change looks fine; needs testing and review from the balance team.

PR #2213, opened 2024-01-25
Add Smash 64 mode

jobrien97:
Adds Smash 64 mode as a custom game mode!

Changes:
- Higher hitstun
- Higher hitlag
- Higher shieldstun
- No DI
- No airdodges
- No walltechs
- Auto Z-canceling
- Lower fall speeds/gravity

2024012421095700-0E7DF678130F4F0FA2C88AE72B47AFDF

### DIFF
--- a/fighters/common/src/function_hooks/change_status.rs
+++ b/fighters/common/src/function_hooks/change_status.rs
@@ -1,5 +1,7 @@
 use super::*;
 use globals::*;
+use utils::game_modes::CustomMode;
+
 
 #[skyline::hook(replace=StatusModule::change_status_request)]
 unsafe fn change_status_request_hook(boma: &mut BattleObjectModuleAccessor, status_kind: i32, arg3: bool) -> u64 {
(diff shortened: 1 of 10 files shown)


fighters/common/src/function_hooks/change_status.rs
```diff
@@ -53,10 +55,27 @@ unsafe fn change_status_request_from_script_hook(boma: &mut BattleObjectModuleAc
     let mut clear_buffer = arg3;
 
     if boma.is_fighter() {
+        match utils::game_modes::get_custom_mode() {
```
WuBoytH, line 58: Instead of putting this in a status change request hook, please put this in is_enable_transition_term instead

WuBoytH (dismissed):

fighters/common/src/function_hooks/change_status.rs
blu-dev: There are places where we transition into airdodge (like wavedash) where we don't check transition term, I think this is fine.

WuBoytH: in that case we really should be refactoring that so that it does use the system, but I won't push too hard for it now

PR #2230, opened 2024-01-29
DJC hotfix's fix pr of the fix pr's fix pr

jaccob101:
sorry (forces dj aerials to inherit the movement arc before allowing cancels)
accidentally undid it oops
[peachassets-1.zip](https://github.com/HDR-Development/HewDraw-Remix/files/14107610/peachassets-1.zip)
fixes (from first impl):
-unnecessary hitstop check during exec
-delaying check until frame 2 of aerial
-peach dj frames (changed bc i failed to realise the above issue)
note for what this does exactly:
-force djc to experience at least 1f of dj arc
-checks for cancel AFTER 2nd motion is added
-it doesn't run the check during hitstop

this combined makes it perform more accurately, but still loses more height with tapjump. no issues were noticed on other characters but could be limited to only ness/lucas if they arise


WuBoytH:
"Who are you?"
"I'm you."
"No, I'm me!"
"I'm me, he says."

PR #2231, opened 2024-01-29
Sakurai angle Melee behavior

jobrien97:
Changes the behavior of the **grounded** Sakurai angle to match Melee, where below 32 units of knockback, your launch angle is 0, while above 32.1 units of knockback, your launch angle is 44.

Differences between games' behavior shown here (HDR currently uses Ultimate's behavior):

image


WuBoytH:
Not being considered at this time.

PR #2249, opened 2024-02-07
Fox Shine Adjustment

WuBoytH:
Made exclusively because Jut asked me to.

# Changelog

### Down Special
- [-] Removes Electric property
- [-] Reduces hitstun by 5 when hitting grounded opponents with the ground version


TeamPearson:
Re added falcos electric properties and reduced fox shine hitstun by 3 frames to match melees values

TeamPearson:
wolfs shine gives much less stun than p+ so I have also reverted the removal of electric attribute for him as well

PR #2250, opened 2024-02-08
WuBoy Misc 2

WuBoytH:
# Changelog

### Side Special - Dancing Blade
- [*] Fixed a bug where ground Side B 1 didn't have a hitbox

### Up Air
- [*] Fixed rumble

### Neutral Special - Homing Attack/Blast Attack
- [*] Fixed an oversight where its FAF was 7 frames earlier than intended.

### Neutral Special - Mega Buster
- [*] Fixed an issue introduced in the previous WuBoy Misc PR that causes Mega Man to have a particularly long jump squat during Mega Buster.

### Side Special - Metal Blade
- [*] Fixed an issue that causes Metal Blade to not despawn when getting hit while throwing it, making it unusable for the rest of the stock.


CrossSockDance:
Would it be too out of scope to request down angled Alolan Whip becoming a grab instead of a hitgrab?

WuBoytH:
> Would it be too out of scope to request down angled Alolan Whip becoming a grab instead of a hitgrab?

Very.

PR #2251, opened 2024-02-08
Mario/Doc ftilt animation adjustment

robertyur:
[mariosassets.zip](https://github.com/HDR-Development/HewDraw-Remix/files/14203084/mariosassets.zip)
Makes it so both Marios no longer appear to be floating when doing any variant of ftilt by a ledge. This slightly reduces their range.

_Before:_

_After:_

_Before:_

_After:_

_Before:_

_After:_


WuBoytH:
Not being considered at this time (he's basicaly gameing jesus let him float)

PR #2309, opened 2024-03-10
Restore Kill Screen

moklmaru:
### **DISCLAIMER: This PR is still in its testing phase and is largely experimental while we iron out the imperfections.**
___
Assets:  [kill-assets-v2.zip](https://github.com/HDR-Development/HewDraw-Remix/files/14579744/kill-assets-v2.zip)

This PR adds a kill effect similar to what is seen in vanilla Ultimate, which was previously removed from HDR due to inconsistencies in knockback calculation.

https://github.com/HDR-Development/HewDraw-Remix/assets/16338870/a3345181-8954-47cf-b0dd-0d0a0579991c

## Differences compared to vanilla:
- New effect coloring on both the kill sparks and finishing background (Not final)
  - if you don't want these or have an existing mod for them, ignore the included .eff file
- The slowdown does not restrict DI or SDI
- The camera will not zoom
- Fewer slowdown frames (~35)
- Different sound effect, being the sound that plays when downing a boss
- The stage and UI do not disappear during the cut-in (however, the stage's model will be blacked out)

## Kill screen conditions:
- The game will check 10 angles across the entire DI range. If at least 9 of them result in a kill, the screen will trigger. This means false finishes can still happen on perfect DI, but are _far_ less common.
- The screen will not trigger if the character will be within 2 units of the ground during their calculated trajectory, if their launch angle is any more horizontal than 30°. This helps save from triggering the screen in certain tech situations.
- The screen will only trigger when the defender is on their last stock*
  - However, this effect can also be turned on and off in training mode. Press attack + shield while holding up taunt to turn them on (indicated by green sparks), and while holding down taunt to turn them off (indicated by a puff of smoke). They are off by default.

To put the angle check into perspective, divert your attention to this diagram.

wahoo

Let's say that Super Mario here hits you with an attack that sends at this *exact* knockback distance. The blue bounding box represents the blast zone, and the red line is the trajectory with no DI. The dotted lines represent the 36° DI window of HDR split into equal segments, which create the 10 trajectories that the game will analyze. In this case, 9 of the 10 checked angles will kill, which will prompt the kill screen to trigger. Any fewer, and it will not. This means that absolute perfect DI in this exact situation can allow for a false finish, but that is in the context of this exact positioning with that exact knockback distance. You have to play with it to get the full picture, but survivals should be very few and far between. Here's a couple more examples with slightly altered trajectories that immediately push it out of "false finish" territory.

10/10 -> Can not be survived | 8/10 -> Will not play kill screen
-- | --
</img> | </img>

Knockback trajectory isn't completely linear like this, so this isn't *actually* completely accurate to the in-game equivalent, but the general idea of what it's doing is largely the same.

TL;DR - it's pretty accurate

## Special / Critical Zoom Moves
This PR also adjusts moves with "special zoom" properties, which is another aesthetic feature added by Ultimate. To match the style of the new kill screen, the duration in which they freeze the game has been standardized to 30 frames with 8x slowdown, and the camera will no longer zoom in.

List of affected characters:
- Donkey Kong: Giant Punch (fully charged)
- Kirby: Hammer Flip (fully charged)
  - All neutral special zoom adjustments are also applied to Kirby's copy abilities
- Luigi: Super Jump Punch (sweetspot)
- Captain Falcon: Falcon Punch
- Jigglypuff: Rest
- Marth: Shield Breaker (on successful shield break)
- Roy: Flare Blade (charged)
- Mr. Game & Watch: Judge (9)
- Wario: Waft (charged)
- Ike: Eruption (charged)
- King Dedede: Jet Hammer (fully charged)
- Lucario: Force Palm**
  - Critical zoom attributes have been removed from this move entirely
- Little Mac: K.O. Punch (charged)
- Mii Brawler: Explosive Side Kick
- Lucina: Shield Breaker (on successful shield break)
- Cloud: Limit Cross Slash / Climhazzard / Finishing Touch
- Ridley: Skewer (sweetspot)
- Chrom: Flare Blade (charged)
- Incineroar: Revenge-boosted attacks
- Hero: Critical hits, Whack, Thwack
- Byleth: Aymr
- Steve: TNT
- Kazuya: Forward smash (sweetspot), Rage Drive

## Known issues:
- May yet be some inaccuracies with SDI/tech situations. Needs more testing.
- Certain moves may result in triggering the kill flash multiple times
- Nana will invoke the kill screen when being knocked out on her last stock, even if the attack doesn't kill Popo

Feel free to play around with it. There are likely still some oversights and inconsistencies in behavior to discover, so this will be polished up over time. Big thanks to blujay for the KB calcs and general assistance with implementation.


utils/src/modules/meter.rs
```diff
@@ -397,33 +399,220 @@ unsafe fn shield_module_send_shield_attack_collision_event(shield_module: *mut u
...
+/// 0x4c -> hitstop frame
+
+const HANDLE: i32 = 0x01FF;
+const COUNTER: i32 = 0x01FE;
+
+pub unsafe extern "C" fn calculate_finishing_hit(defender: u32, attacker: u32, knockback_info: *const f32) {
+    *(knockback_info.add(0x4c / 4) as *mut u32) = 60;
+    let defender_boma = &mut *(*utils_dyn::util::get_battle_object_from_id(defender)).module_accessor;
+    let attacker_boma = &mut *(*utils_dyn::util::get_battle_object_from_id(attacker)).module_accessor;
+    if !defender_boma.is_fighter() && !defender_boma.is_weapon() { return; }
```
techyCoder81: Should be `if !defender_boma.is_fighter()` only? Do you want this to run for weapons?

PR #2311, opened 2024-03-11
Lucina Down Special Follow Up

CSharpM7:
https://github.com/HDR-Development/HewDraw-Remix/assets/13909643/5b2b4602-86e7-487e-855f-5350693ec0ef

- Changes Lucina's Down Special from the counter animation to a dash. Similar to previous iterations, it is once per airtime and can be canceled using Special + Directional Input. Up, Side, and Down use the final hits of Dancing Blade
- In the air, Low will move Lucina downwards but will stall her in the air upon hitting an opponent
Known issues:
- Effect/Sound acmd will attempt to run again if changing situation before frame ~15


WuBoytH:
Not being considered because what the hell is lucina right now anyway we gotta talk about it

PR #2333, opened 2024-03-21
Lower dash end speed multiplier

jobrien97:
0.225 -> 0.0

The speed multiplier when performing the following actions out of dash:
- Dashback
- Pivot
- Grounded normals
- Grounded specials
- Item throws

is currently 0.225 to emulate Melee's behavior. Upon further consideration, there is no reason why this value shouldn't simply be 0, halting all momentum out of dash. This makes for slightly snappier dash dancing, and further differentiates performing actions out of dash vs. run.


SuddyN:
feels kinda weird, makes dashdancing pretty strong. I could see a case for it but I'm not sure if I like it? Maybe we should try a value in the middle

WuBoytH:

PR #2356, opened 2024-04-02
B input rapid jab finishers changed to gentlemen

zrksyd:
Affects
- King Dedede
- Falco
- Fox
- Mr. Game and Watch
- Kirby
- Bowser Jr.
- Mewtwo
- Mii Brawler
- Villager (B input removed)
- Palutena
- Ivysaur
- Sheik
- Simon
- Zelda

Timing for when gentlemen should be activated may need adjustments.

### DIFF
--- a/fighters/common/src/opff/mod.rs
+++ b/fighters/common/src/opff/mod.rs
@@ -113,7 +113,7 @@ pub unsafe fn moveset_edits(fighter: &mut L2CFighterCommon, info: &FrameInfo) {
     cancels::run(boma, info.cat, info.status_kind, info.situation_kind, info.fighter_kind, info.stick_x, info.stick_y, info.facing);
     ledges::run(fighter, boma, info.cat, info.status_kind, info.situation_kind, info.fighter_kind, info.stick_x, info.stick_y, info.facing);
     var_resets::run(boma, info.cat, info.status_kind, info.situation_kind, info.fighter_kind, info.stick_x, info.stick_y, info.facing);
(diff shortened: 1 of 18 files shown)


SkewedAskew (changes requested):
- Near-universally, the input window to transition to rapid jab persists for too long, making gentlemans feel very clunky when not hitting opponents. Need to check the exact frames and match closer to vanilla characters
- Jab -> finisher transition frames in most cases feel like they need to be pushed back a small amount
- These should also be made into separate acmd scripts with earlier cancel frames and less kb (and in most cases smaller hitboxes)

Dedede - decrease hitbox vertically
Falco - no other issues
Fox - no other issues
Gnw - no way to actually get rapid jab now
Kirby - decrease hitbox horizontally
Bowser Jr. - decrease hitbox horizontally, animation needs adjustment (starts with other hand already extended)
Mewtwo - same issue as Gnw, decrease hitbox horizontally
Mii Brawler - decrease hitbox horizontally
Villager - as I understand, this should just be removed entirely?
Palutena - transitions feel ok but why is it so massive
Ivysaur - decrease hitbox both horizontally and vertically, but transitions feel perfect
Sheik - decrease hitbox slightly
Simon - redo hitbox entirely, jab 2 transition occurs immediately, hold-to-gentleman is impossible due to held whip on jab1 (consider removing)
Zelda - transitions feel fine due to jab1's multiple hits, decrease hitbox horizontally

WuBoytH:
Will be remade and included in a future PR

PR #2377, opened 2024-04-23
Greninja dair on hit changes

EluTran:
In Smash 4, Greninja's dair on hit had a FAF of 13 frames. This was a core aspect of his kit in Smash 4, as both sourspot and sweetspot dair hit confirm combos led to many creative combo routes, and as well as leaned into his themes of being a frog. Ultimate would go on to nerf Greninja's on hit dair FAF, making it significantly harder to combo out of, and less rewarding to hit an already-risky move to commit to.

This PR aims to adjust Greninja's dair by decreasing the on hit dair FAF back to its original 13 frames like in Smash 4, making it more rewarding to land. To compensate for this change, due to dair becoming stronger on hit, the landing lag has also been increased to discourage haphazard use and to punish users for whiffing the move. The move's bounce Y speed multiplier has also been lowered. This makes Greninja bounce less high on hit, letting him stay closer to the ground post-bounce, while also punishing him on block too.  In addition, the sourspot's angle and base knockback have been adjusted to better lend itself to the lower bounce and combos from the sourspot dair.

**CHANGELOG**
**DAIR**
- [+] FAF on hit: 17F -> 15F
- [-] Landing Lag: 16F -> 20F
- [/] Bounce Speed Y Multiplier: 0.75 -> 0.57 
- [/] Sourspot angle: 75 -> 70
- [/] Sourspot BKB: 75 -> 60
- [-] Sourspot Shieldstun Multiplier: 1.0x -> 0.5x

### DIFF
--- a/fighters/gekkouga/src/acmd/aerials.rs
+++ b/fighters/gekkouga/src/acmd/aerials.rs
@@ -210,7 +210,7 @@ unsafe extern "C" fn game_attackairlw(agent: &mut L2CAgentBase) {
     }
     frame(lua_state, 20.0);
     if is_excute(agent) {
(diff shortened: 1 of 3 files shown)


WuBoytH:
It might be best if down air launched opponents less high and if Greninja didn't bounce as high while using it. This buffs his on-hit by letting him stay closer to the ground post-bounce, while also punishing him on block too.

EluTran:
Thank you for the feedback, I've made some more adjustments. First off, figured out how to edit the properties of the bounce's height, so I've adjusted that so that it's now lower. The sourspot's angle and BKB have also been adjusted to both account for the lower bounce, as well as lend itself better to following up off of a sour dair.

Comparison
Beta:
image

Current commit: 
image

- Thanks!

PR #2385, opened 2024-04-30
Change Shadow Sneak Smash Attack Cancel to A Jump Cancel

EluTran:
Self-explanatory. Gets rid of the weird round-about-ness that currently exists with shadow sneak where people would smash attack and then jump cancel the smash attack. Now it's simply a jump cancel input, like Ike's side b.

https://github.com/HDR-Development/HewDraw-Remix/assets/167844127/8d99a1b0-9264-486e-a5ce-64975c45a0f6

### DIFF
--- a/fighters/gekkouga/src/opff.rs
+++ b/fighters/gekkouga/src/opff.rs
@@ -11,20 +11,12 @@ unsafe fn max_water_shuriken_dc(boma: &mut BattleObjectModuleAccessor, status_ki
     }
 }
 
-// Greninja Shadow Sneak Smash Attack Cancel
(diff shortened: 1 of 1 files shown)


EluTran:
New commit now requires Greninja to be on the ground for at least 1 frame before being able to cancel the shadow sneak. This pushes the ability to cancel SS to one frame later. So, technically a miniscule nerf. This also resolves issues with the smoke's appearance, as previously the move was able to be cancelled before the smoke appeared.

The actual window of execution remains the same at 5 frames.

https://github.com/HDR-Development/HewDraw-Remix/assets/167844127/270cfc8b-855b-43ed-8bd4-c1982df54927

EluTran:
Looking forward to a response on this PR :)

SuddyN:
not planned at this time

PR #2391, opened 2024-05-04
Zelda Tweaks + Teleports

jaccob101:
Makes some things about Zelda cleaner (imo), minor buffs and a couple teleport changes.
Assets: [dtaunt pr.zip](https://github.com/HDR-Development/HewDraw-Remix/files/15210337/dtaunt.pr.zip)
### General:
- [*] Teleport wallrides now preserve horizontal momentum if no longer touching wall
Affects Zelda, Sheik, Mewtwo, and Palutena. Wall-touch check changed because it wasn't working on training stage.
### Up Special (Vanish):
- [$] Re-appearance changed to no longer have a flash effect, just a smoke effect. This better communicates that there's no hitbox. 
### Jab:
- [$] Effects adjusted to be more simple and closer match hitbox
### Rapid Jab:
- [+] Damage: 0.4 -> 0.6
- [+] BKB: 7 -> 5
### Jab Finisher:
- [-] BKB: 55 -> 40
- [/] Hitlag Multiplier: 2x -> 1.2x
### Down Tilt:
- [$] Effects pulled in to match hitbox
- [+] Hitbox activity: 6-10 -> 5-9
Couldn't find the startup nerf in changelog and it wasn't done via motion rating so visuals were 1f off
### Forward Aerial:
- [-] Hitbox Activity: 8-11/12-16 -> 8-11/12-14
- [-] Damage: 9/8 -> 8
- [-] Angle: 38 -> 361
- [-] BKB: 25 -> 10
- [-] KBG: 89/100 -> 95
- [+] FAF: 44 -> 41
### Back Aerial:
- [-] Hitbox Activity: 6-9/10-14 -> 6-9/10-12
- [-] Damage: 10/7.5 -> 8
- [-] Angle: 55 -> 361
- [-] BKB: 10/65 -> 10
- [-] KBG: 103/50 -> 95
This makes sour kicks match P+ but with 10 more kb at all percents (~10% more down from roughly 20% more kb). Angles/dmg normalized for sour hits
### Down Aerial:
- [$] Ending motion slowed down
### Side Special (Din's Fire):
- [+] Dins recast despawns old dins upon placement of new dins mine, rather than frame side special startup.
- [$] Flash effect scales more dynamically, matched to hitbox size a bit better
- [$] SFX added 20 frames before explosion
### Up Special (Farore's Wind):
- [!] Reappearance now snaps to platforms
- [!] Grounded re-appearance now reversible
- [$] Effects shrunk on both startup and re-appearance
- [*] Buffered specials on frame 0 slide-off removed
- [-] Air stall end frame: 12f -> 9f (aerial teleport should be less floaty)
- [-] Hitbox sizes: 13/8, 12.5/7.5 -> 11.2/7.5 (re appearance)
- [*] FAF corrected by fixing cancel frame in motion rate formula. 1f extra lag than intended
### Down Special (Phantom):
- [*] Effects no longer play during win screen / game start, logic for added mechanics should be improved and more legible
- [*] Platdrop mechanic now requires a fresh stick flick, no longer gated to frame 10
- [-] Angle: 6 -> 25 (aerial max charge windbox)
- [/] Grounded cancel position: 19u -> 18u in front of Zelda (matches non-cancel ver)
### Forward Throw:
- [+] Release frame: 30 -> 27 (less frame adv but better for dins traps)
### Back Throw:
- [+] Release frame: 31 -> 27 (rev)
- [+] FAF: 51 -> 48 (rev)
### Down Taunt:
- [$] Now plays magic sfx

Hitboxes:
Up special reappearance:
Before:
image
After:
image


Z-hime:
Good PR! 

**[UpB]**
-Teleport fixes overall are very good.
-Makes the directional inputs more accurate and platform snap finally implemented is great.
-B turnaround adds some more mix up potential to the move, and gives her more opportunities to use her grab game. Probably one of the standout features of the PR.
-GFX indicator on teleport seems to be a bit large. Overall, I'm not really a fan of these types of "indicators" but if the rest of the cast is getting them, I guess it's okay. Would just reduce the size and maybe match the color of teleport so that it's less visually intrusive when using cancels for movement -- could actually become more distracting than actually aiding to the opponent.

**[Jab]**
-Jab changes are good
-Reduced hitlag is welcome on the launch hit, spacing is slightly better for short phantom cancel tech chase scenarios.

**[Dtilt]**
-Good, nothing really stands out. Seems to be fine change moving the active frames.

**[Down B]**
-Phantom Windbox change also very good as it was a lot of the time pushing when it really probably shouldn't have.
Was annoying for the opponent but now will actually function when used intentionally. Reposition on phantom cancel also good visually to match the regular grounded distance since previously adjusted.

**[Throws]**
-Both release frames feel more natural for trapping and acting afterwards

**[Dins]**
-Didn't notice too much of a difference in the spawn point, but makes sense visually

Taunt sound - Definitely. <3

PR #2428, opened 2024-05-27
Implement 'Rivals 2'-Style Shield Breaks

stingers135:
https://github.com/HDR-Development/HewDraw-Remix/assets/115596132/6f2e57a1-dc09-496b-93f1-4da2d5f02fb6

Shield breaks being such a strong punish with the base system has led to many issues with shield damage throughout the games history. Implementing this system opens design space by making shield damage a more legitimate balance lever compared to how it is currently.


fighters/common/src/general_statuses/shield/shield_break_down/main.rs
```diff
@@ -0,0 +1,55 @@
+// status imports
+use super::*;
+use globals::*;
+
+#[skyline::hook(replace = smash::lua2cpp::L2CFighterCommon_status_ShieldBreakFly_Main)]
```
WuBoytH: you hook this status twice, I assume on accident.

WuBoytH (changes requested):

SuddyN:
Considered but replaced with a different shield break rework

PR #2435, opened 2024-06-09
Copy Abilities Strike Again + Small Adjustments

ethangarsia517:
Hm yes, new Neutral Special. Note that all Neutral Special changes are new from Beta.

General:
~ [+] Probability of losing Copy Ability in a single hit reduced by 33.3%. It's now equal to ⅔ of damage taken in a single hit.
~ Example: if Kirby gets hit by a move that deals 15%, the probability of losing copy ability on that hit is 10%, where it would previously be 15%.

**Copy Abilities: ---------------------------------------**
~ Roy (Flare Blade):
   ~ [!] Correctly repositioned hitboxes to the left arm and sword (because wtf was going on)
Before: 
roywtf

After: 
royfixed

~ Lucas (Offense Up):
   ~ [!] Fixed an issue where the sound for charging Offense Up was not playing
   ~ [!] Fixed an issue where Kirby would continue to have the PSI effects if the Copy Ability was lost

~ Ridley (Plasma Breath):
   ~ [!] Fixed an issue where Kirby would T-Pose when charging the move if Kirby changed from air to ground, or ground to air, which consequently also allowed it to charge infinitely.

~ Dark Samus (Charge Shot):
   ~ [!] Fixed an issue where if you pressed Shield in the air to cancel Charge Shot charging, Kirby would transition to airdodge instead of normal fall.

~ Banjo & Kazooie (Blue Eggs):
   ~ [!] Disabled the ability for Kirby to perform Forward Tilt out of Blue Eggs.
   ~ [$] Added sweat effects for when Kazooie's Blue Egg Power is decreasing.

~ Palutena (Colorless Attack & Auto Reticle)
   ~ [R] Reworked to an Energy System. Kirby gains 1 Cyan Energy point when he lands an attack with any of his normals (except Neutral Air, Jab, and Dash Attack), and 3 Cyan Energy points when he lands Colorless Attack, which still comes out on Frame 7.
   ~ [R] On frame 14 of Colorless Attack, if Kirby has at least 3 energy points, he will proceed to use Auto Reticle. The bullet consumes 1 energy point when it spawns. Since Kirby gains 3 on hit, this will always result in Auto Reticle and consume that gained energy if a target is found.
   ~ [R] Kirby can hold up to 6 Cyan Energy points, and can perform Auto Reticle twice. If Colorless Attack connects when Kirby has 4 or more energy points, Kirby will be able to use that extra energy. If it is not consumed, it will be lost and reset to 6.
   ~ [-] FAF: 33 -> 43
   ~ [-] Can no longer be dash or jump Canceled on hit
   ~ [!] Modified Auto Reticle Stats
https://discord.com/channels/659964948365049887/660220440370872332/1249402049792053360

**General Adjustments: --------------------------------**
Hammer Flip (Side Special):
~ Air:
   ~ [+] First frame able to fastfall/aerial drift: F12 -> F1
   ~ [+] Can now aerial drift when vertical speed is 0 or positive
   ~ Second Swing:
      ~ [-] Angle: 35 -> 38
      ~ [/] KBG: 93 -> 94

Stone (Down Special):
~ [-] Minimum frames in Stone: 10F -> 13F
~ Air:
   ~ [+] BKB: 69 -> 75
   ~ [+] KBG: 75 -> 79
   ~ [+] Shield Damage: 0 -> 3
~ Ground:
   ~ ~~[!] Added ground-only windbox on both sides~~
   ~ [+] Shield Damage: 0 -> 13
~ Note: Shield Damage for both Air and Ground variants were supposed to be changed to 10 in a previous nightly, but we're actually still 0 due to an error.

Down Smash:
~ Middle Hitbox:
   ~ ~~[/] Decreased priority. Inner leg hitboxes now prioritized.~~
   ~ [+] Damage: 10.0% -> 11.0%

Down Air:
~ [+] Land Auto Cancel: F42 -> F40
~ ~~[+] FAF: F46 -> F44~~

Up Throw:
~ [+] KBG: 68 -> 70


MythicalWorks:
Tried it. Fixes the mentioned issue with Ridley's copy ability and improves overall polish.

WuBoytH:
Most of this looks good, but please revert the following:
- down smash priority change
- stone windbox
- down air (faf only, autocancel is fine)

PR #2448, opened 2024-06-24
Global aerial FAF ledgegrab

jobrien97:
Allows all aerials to grab ledge once their FAF is reached, rather than when their animation completes. The difference this makes will largely vary between each character and move, as it depends on how many frames exist between the move's FAF and the end of its animation.


SkewedAskew:

PR #2475, opened 2024-07-31
Uncapped G2A Floats

jaccob101:
Makes floats have uncapped motion transfer the same way aerials do, rather than instantly dropping to the air speed cap. Mostly affects Daisy but has been done to all floats.

### DIFF
--- a/fighters/daisy/src/status/uniq_float.rs
+++ b/fighters/daisy/src/status/uniq_float.rs
@@ -3,8 +3,15 @@ use super::*;
 // FIGHTER_PEACH_STATUS_KIND_UNIQ_FLOAT_START
 
 extern "Rust" {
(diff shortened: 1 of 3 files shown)


fighters/daisy/src/status/uniq_float.rs
```diff
@@ -2,6 +2,15 @@ use super::*;
 
 // FIGHTER_PEACH_STATUS_KIND_UNIQ_FLOAT_START
 
+extern "Rust" {
```
WuBoytH, line 5: please combine this extern "Rust" with the one below for cleanliness.

WuBoytH (changes requested):

WuBoytH (approved):

SkewedAskew:

PR #2488, opened 2024-08-02
Bayonetta Fixes / Neutral B Slowdown Adj

jaccob101:
### General:
- [*] Transitions from side b kick and neutral b cancel now use the correct idle anim
- [*] Neutral b cancel and side b slide-offs now added to fastfall opff
### Neutral Special
- [/] Air_start_mul_y: 0.25 -> 0.35 (looks better)
- [$] Special lag portion now slows down during the initial portion rather than slowing the whole anim
- [$] Landing during cancel now transitions into landing anims
- [*] Special lag now always applied, currently scenarios where it can get cut off or ignored (like landing during cancel anim)


WuBoytH:
Bro.

PR #2511, opened 2024-09-04
Enable hitfalling for King K Rool

stingers135:
_This idea came to me in a dream_

King K Rool's kit intuitively seems like it would work well with hitfalling, mainly due to how his Uair works. Giving it additional utility by letting it serve better as a combo tool instead of how it works now as only a finisher should open up the character more. His other aerials should benefit as well, but that was the main inspiration.

He is generally regarded as a bottom 3 character, as well as relatively inflexible in playstyle. This seems like it could be an interesting way to open him up and make him more viable at the same time.


SkewedAskew:
Not being considered. Krool has been evaluated and changes have been made for the next release

PR #2515, opened 2024-09-09
Allow most custom modes to run in local wireless multiplayer

BrianAllred:
This change allows all custom game modes except for Smash Tag to run in local wireless multiplayer. Unfortunately, Smash Tag causes a panic when transitioning from the waiting room to the match.

To trigger custom game modes in local wireless, **the host** (player who opens the room) must **hold R** when opening the room (similar to holding R when selecting a rule set in regular local play). **Anyone who joins** the room must press **DPad Down** to select the same mode(s) **before joining** the room (similar to pressing DPad Down in an online arena).


BrianAllred:
If anyone wants to help fix Smash Tag:

The current issue is that the Smashball aura effect begins in the waiting room, then the entity with the effect is despawned during actual match loading.

PR #2531, opened 2024-10-31
Copy Abilities Adjustments

ethangarsia517:
Lucario -
- Fixes a crash that happened when shooting a fully charged Aura Sphere

Chrom -
- Repositioned hitboxes to be on the sword instead of on his feet.

Piranha Plant -
- Added indicator for Adaptive Roots. Note that the stance can still only be changed by discarding the copy ability, and copying Piranha Plant again while she is in another stance.

Sora -
- Added 35 frame lockout for spells.
- Adjusted Blizzaga to work and behave exactly like Sora's. This copy ability now has no differences to the original.


fighters/kirby/src/status/mod.rs
```diff
@@ -182,6 +182,14 @@ unsafe extern "C" fn ganon_should_use_special_n_callback(fighter: &mut L2CFighte
     }
 }
 
+unsafe extern "C" fn trail_should_use_special_n_callback(fighter: &mut L2CFighterCommon) -> L2CValue {
+    if VarModule::is_flag(fighter.battle_object, vars::trail::instance::DISABLE_SPECIAL_N) {
```
WuBoytH: This should be behind a copy ability check

WuBoytH (changes requested):

fighters/kirby/src/opff/copy.rs
```diff
@@ -475,6 +495,29 @@ unsafe fn trail_magic_cycle(fighter: &mut L2CFighterCommon, boma: &mut BattleObj
     }
 }
 
+// handles the speed and disappearance of blizzaga effects
+unsafe fn trail_flower_frame(boma: &mut BattleObjectModuleAccessor) {
```
WuBoytH, line 499: this should be behind a copy ability check as well

WuBoytH (changes requested):

PR #2586, opened 2025-07-22
Robot 3.0 - Robotic Uprising

stingers135:
### General
```diff
+ [+] Maximum Runspeed 1.618 -> 1.66
```

### Dash Attack
```diff
Early:
~ [/] BKB: 47 -> 55
+ [+] FAF: 36 -> 34
```

Dials back some of the previous nerfs, but does not fully revert. Pre-nerf, the FAF was 32 and the old BKB was 70, for reference.

### Forward Air
```diff
Late:
+ [+] BKB 35 -> 40
````

Intended to make weak Fair slightly less punishable on-hit, but still should be unsafe, particularly on grounded opponents.

### Neutral Air
```diff
+ [+] KBG 100 -> 105
```

Neutral Air was intended to be ROBs primary kill option in the air since Uair is no longer used for that purpose, but it was falling short. Slight buff to help it do it's intended job better.

### Dash Item Throw
```diff
~ [/] Release Frame: 4 -> 6
```

This will make it possible to DITCIT and choose the direction you throw the projectile instead of it being forced to be thrown forwards, but will impact standard dash item toss by making it slower.

### Up Special
```diff
~ [/] Slightly reworked the angle that UpB sends at depending on charge level/angle. Should feel smoother overall.
~ [/] Minimum Launch Distance 0 -> 0.05

Ground:
+ [+] Rotation Speed: 1.0 -> 1.5
+ [+] Charge Speed: 1.0 -> 2.0
- [-] Maximum Charge Length - 0.5x
```

The first change about minimum launch distance is intended to change the behavior where instant up-b usages offstage would lose height, which felt janky. Should now stall roughly in place which feels more natural. 

Grounded upspecial was intended to be able to be used aggressively with a lower FAF compared to aerial, but it took too long to charge for its intended purpose. Adding this to incentivze its usage more in neutral. However, now you cannot hold the charge as long, which will impact your ability to fish for spikes at ledge if you remain onstage.


fighters/robot/src/acmd/specials.rs
```diff
@@ -338,9 +364,17 @@ unsafe extern "C" fn game_specialhirise(agent: &mut L2CAgentBase) {
...
     frame(lua_state, 1.0);
     if is_excute(agent) {
-        let mut damage = charge_frame / 3.0;
+        
+        if boma.is_prev_situation(*SITUATION_KIND_GROUND) { 
+            let mut damage = charge_frame / 1.5;
+        } else {
+            let mut damage = charge_frame / 3.0;
+        }
+
```
SuddyN: builds fail because `damage` is no longer in-scope of the attack functions below

```rs
let damage = if boma.is_prev_situation(*SITUATION_KIND_GROUND) { 
    charge_frame / 1.5
} else {
    charge_frame / 3.0
};
```

fighters/robot/src/status/special_hi.rs
```diff
@@ -109,34 +109,84 @@ unsafe extern "C" fn special_hi_main_loop(fighter: &mut L2CFighterCommon) -> L2C
     let max_fuel = fighter.get_param_float("param_special_hi", "energy_max_frame");
     let fuel_increment = 2.0; // how much fuel is consumed by the charge per frame
     let min_cost = 20.0; // minimum amount of fuel consumed on use
-    let required_fuel = (fuel_increment * charge_frame).clamp(min_cost, max_fuel);
+    let required_fuel = if fighter.is_situation(*SITUATION_KIND_AIR) { (fuel_increment * charge_frame).clamp(min_cost, max_fuel) } else { ((fuel_increment * 2.0) * charge_frame).clamp(min_cost, max_fuel)};
```
SuddyN: instead of duplicating the logic here you should really just multiply fuel_increment before line 112

fighters/robot/src/status/special_hi.rs
```diff
@@ -109,34 +109,84 @@ unsafe extern "C" fn special_hi_main_loop(fighter: &mut L2CFighterCommon) -> L2C
...
+            let direction = fighter.lr() * reverse; // determines the direction to rotate
+            let angle = (rot_x + (rot_amount * direction)).clamp(-60.0, 60.0);
+            PostureModule::set_rot(fighter.module_accessor, &Vector3f::new(angle * 0.3 * fighter.lr(), 0.0, 0.0), 0);
+            VarModule::set_float(fighter.battle_object, SPECIAL_HI_ROT_X, angle);
+
+            // changes direction if rotation crosses center threshold
+            if rot_x == 0.0 && fighter.is_stick_backward() {
+                PostureModule::reverse_lr(fighter.module_accessor);
+                PostureModule::update_rot_y_lr(fighter.module_accessor);
+            }
```
SuddyN: more unnecessarily duplicated code

fighters/robot/src/status/special_hi.rs
```diff
@@ -173,12 +229,19 @@ unsafe extern "C" fn special_hi_main_loop(fighter: &mut L2CFighterCommon) -> L2C
...
 
     if charge_frame >= 10.0 {
-        launch_speed.y = ((1.65 + (0.05 * charge_frame)) - (0.025 * rot_x.abs())).min(3.75);
+        if fighter.is_situation(*SITUATION_KIND_GROUND) {
+            launch_speed.x = 0.30 * rot_x.abs() * (((charge_frame * 2.0) - 18.0).clamp(0.0, 32.0) / 32.0);
+            launch_speed.y = ((1.65 + (0.10 * charge_frame)) - (0.035 * rot_x.abs())).min(3.75);
+        } else {
+            launch_speed.x = 0.15 * rot_x.abs() * (((charge_frame) - 18.0).clamp(0.0, 32.0) / 32.0);
+            launch_speed.y = ((1.65 + (0.05 * charge_frame)) - (0.025 * rot_x.abs())).min(3.75);
+        }
```
SuddyN: it's genuinely hard to follow what the launch speeds are supposed to be for different scenarios in this larger function. You're doing a lot of math in-line even though it looks like only a few of these numbers actually change in the different if blocks. You should try to extract common patterns out to earlier in the function so they can be reused without bloating the function.

SuddyN (changes requested):
You're effectively just multiplying charge_frame by 2 everywhere right? can't you just do that to the actual value of charge_frame, rather than manipulating the result all over the place?

fighters/robot/src/acmd/specials.rs
```diff
@@ -288,8 +288,13 @@ unsafe extern "C" fn effect_specialslw(agent: &mut L2CAgentBase) {
 unsafe extern "C" fn game_specialhi(agent: &mut L2CAgentBase) {
     let lua_state = agent.lua_state_agent;
     let boma = agent.boma();
+    frame(lua_state, 1.0);
+    if agent.is_situation(*SITUATION_KIND_GROUND) {
+        FT_MOTION_RATE(agent, 4.0/(14.0-1.0));
```
SuddyN:
```suggestion
        FT_MOTION_RATE_RANGE(agent, 1.0, 14.0, 4.0;
```

fighters/robot/src/acmd/specials.rs
```diff
@@ -288,8 +288,13 @@ unsafe extern "C" fn effect_specialslw(agent: &mut L2CAgentBase) {
...
+    }
     frame(lua_state, 14.0);
     if is_excute(agent) {
+        FT_MOTION_RATE(agent, 1.0);
         WorkModule::off_flag(boma, *FIGHTER_STATUS_ATTACK_AIR_FLAG_ENABLE_LANDING);
     }
```
SuddyN:
```suggestion
    frame(lua_state, 14.0);
    FT_MOTION_RATE(agent, 1.0);
    if is_excute(agent) {
        WorkModule::off_flag(boma, *FIGHTER_STATUS_ATTACK_AIR_FLAG_ENABLE_LANDING);
    }
```

fighters/robot/src/acmd/specials.rs
```diff
@@ -310,7 +336,11 @@ unsafe extern "C" fn effect_specialhi(agent: &mut L2CAgentBase) {
             }
             LAST_EFFECT_SET_RATE(agent, 0.75);
         }
-        wait(lua_state, 15.0);
+        wait(lua_state, 10.0);
+    }
+    frame(lua_state, 27.0);
+    if is_excute(agent) {
+        EFFECT_OFF_KIND(agent, Hash40::new("robot_lamp_l"), false, false);
```
SuddyN: Right above this you loop 20 times waiting 10 frames each. That means the animation is at frame 215 here, not 27 as you expect. If the animation is less than 215 frames this code will never be reached

SuddyN (changes requested):
found some more stuff

SuddyN (approved):

SuddyN:
I rewrote your PR description to fit our style guides. Please let me know if anything is now inaccurate.

PR #2608, opened 2025-09-29
Further Kirby Adjustments (for real this time)

ethangarsia517:
# Changelog
### General
```diff
+ [+] Probability of losing Copy Ability: 0.667 -> 0.6
# Probability is ((value * damage taken in a single hit)/100).
```

### Dash Attack
```diff
Late Hit:
- [-] Angle: 65 -> 62
- [-] BKB: 45 -> 50
# Decreased reward for combo potency 
```

### **Back Aerial**
```diff
Early Hit:
- [-] Damage: 14% -> 13%
~ [/] BKB: 15 -> 19
- [-] KBG: 108 -> 104
# Power decreased for 50/50 purposes with Up Aerial, still decent finisher
```

### Inhale
```diff
- [-] Minimum Frames opponent is held: 17 -> 12
- [-] Maximum Frames before grab release: 600 -> 300
```

### Pikmin Toss (Neutral Special)
```diff
+ [+] Reverted Max Pikmin at a time: 2 -> 3
~ [*] Pikmin now properly stale with Neutral Special
```

### Cutter Dash
```diff
Air:
~ [*] Fixed an issue that caused the move to cancel early if used too close to the ground
```

### Hammer Flip
```diff
! [R] All hitboxes when not fully charged can no longer reverse hit
! [!] Changed handle to weak hit with increased priority when not fully charged

Ground:
- [-] Startup increased by 1F
- [-] FAF (Not Fully Charged): 55 -> 60
- [-] FAF (Fully Charged) 60 -> 65

Ground (Weak Hit):
~ Damage: 19 -> 12%
~ BKB: 60 -> 74
~ KBG: 78 -> 66
~ Angle: 48 -> 63
~ Hitlag: 1.1 -> 1.0

Ground (Fully Charged)
- [-] Size (Handle): 3.9 -> 3.6
- [-] Damage: 35% -> 32%
- [-] BKB: 60 -> 55
- [-] KBG: 78 -> 72

Air (Strong Hit 1)
+ [+] Damage: 11% -> 12%
~ [/] BKB: 67 -> 60
- [-] KBG: 106 -> 102

Air (Weak Hit 1)
~ Damage: 12% -> 10%
~ BKB: 67 -> 66
~ KBG: 106 -> 73
~ Angle: 50 -> 65

Air (Strong Hit 2) -
- [-] KBG: 94 -> 92

Air (Weak Hit 2) -
~ Damage: 10% -> 8.5%
~ BKB: 67 -> 66
~ KBG: 94 -> 70
~ Angle: 38 -> 53

# Albeit having decreased priority, strong hits are still easy to hit but may require better spacing. With enough and damage charge even the weak hits can KO.
```

### Stone
```diff
- [-] Changed to Once Per Airtime, including when initially used grounded.
! [!] Gravity and air drift is enabled on Frame 14 when transforming out of Stone.
- [-] Heavy Armor (Stone): 25% -> 15%
+ [+] Knockback Multiplier when Stone Heavy Armor is broken: 1.0 -> 0.5

Ground:
+ [+] Startup: 10F -> 9F
- [-] Maximum Frames in Stone on the ground: 150F -> 90F
+ [+] Shield Damage when started on the ground: 13 -> 16
+ [+] Shield Damage when started in the air: 6 -> 13
 

Air:
~ [/] Fall Speed: 4.5 -> 3.9
- [-] Shield Damage when started in the air: 3 -> 0
- [-] Shield Damage when pushed off an edge: 3 -> 0

# Getting both hits on shield when starting in the air went from 41 -> 45 total. Using it on the ground went from 27 -> 30.
# Stone time is infinite in the air and does not count down until hitting the ground.
# Enabling gravity sooner means Kirby cannot perform an aerial out of Stone if he was grounded. But, it sort of land cancels and transitions nicely into grounded options if not jumped or airdodged out of.
# For now, these changes reduce stalling, encourage challenging Stone on the ground and makes the interaction faster paced, while making it riskier if Kirby is close to an unpunished enemy at the end of the move.
```


dynamic/src/consts.rs
```diff
@@ -823,6 +823,7 @@ pub mod vars {
             //pub const DISABLE_SPECIAL_HI: i32 = 0x01FF; //Weird value to avoid conflicts with copy ability values
             pub const SPECIAL_N_PICKEL_CYCLE_MATERIAL: i32 = 0x01F4;
             pub const SPECIAL_LW_USED_JUMPS: i32 = 0x01F5;
+            pub const DISABLE_STONE: i32 = 0x01F8;
```
SkewedAskew: Small thing, but please rename this to match our variable naming conventions
```rs
pub const DISABLE_SPECIAL_LW: i32 = 0x01F8;
```

fighters/kirby/src/opff.rs
```diff
@@ -55,6 +55,24 @@ unsafe fn inhale_forced_end(fighter: &mut L2CFighterCommon) {
     }
 }
 
+unsafe fn stone_escape_handler(fighter: &mut L2CFighterCommon) {
+    // disables stone if broken out of for any reason
+    if fighter.is_prev_status(*FIGHTER_KIRBY_STATUS_KIND_STONE_STONE)
+    && !VarModule::is_flag(fighter.battle_object, vars::kirby::instance::DISABLE_STONE) {
+        VarModule::on_flag(fighter.battle_object, vars::kirby::instance::DISABLE_STONE);
+    }
```
SkewedAskew: This should be moved to the `End` status of `FIGHTER_KIRBY_STATUS_KIND_STONE_STONE`. Using opff should be avoided unless the alternatives are extremely difficult or involved, as everything in it will run at least once per frame (which is costly if we keep putting more things in it)

For convenience, here is the End status in question, just stick it in `status/special_lw.rs`:
```rs
unsafe extern "C" fn stone_stone_end(fighter: &mut L2CFighterCommon) -> L2CValue {
    VarModule::on_flag(fighter.battle_object, vars::kirby::instance::DISABLE_SPECIAL_LW);
    if !fighter.is_status(*FIGHTER_KIRBY_STATUS_KIND_STONE_END) {
        fighter.change_motion_by_situation("fall", "wait", 0.0, 1.0, false, 0.0, false, false);
    }

    return 0.into();
}

pub fn install(agent: &mut Agent) {
    agent.status(End, *FIGHTER_KIRBY_STATUS_KIND_STONE_STONE, stone_stone_end);
}
```

You shouldn't need to check if the flag is already on as it's going to be set on anyway

fighters/kirby/src/opff.rs
```diff
@@ -55,6 +55,24 @@ unsafe fn inhale_forced_end(fighter: &mut L2CFighterCommon) {
...
+
+    // enables movement on frame 14
+    if fighter.is_status(*FIGHTER_KIRBY_STATUS_KIND_STONE_END)
+    && MotionModule::frame(fighter.module_accessor) >= 14.0 {
+        if fighter.is_situation(*SITUATION_KIND_AIR) {
+            if KineticModule::get_kinetic_type(fighter.module_accessor) != *FIGHTER_KINETIC_TYPE_FALL {
+                KineticModule::change_kinetic(fighter.module_accessor, *FIGHTER_KINETIC_TYPE_FALL);
+            }
+        }
+    }
```
SkewedAskew: Kinetic types don't need to be constantly set/changed, just once. This can be placed in the `game` acmd for `special_lw2` and `special_air_lw2` on frame 14, and also should not need to check the kinetic type beforehand

fighters/kirby/src/status/mod.rs
```diff
@@ -16,6 +16,14 @@ unsafe extern "C" fn should_use_special_hi_callback(fighter: &mut L2CFighterComm
     }
 }
 
+unsafe extern "C" fn should_use_special_lw_callback(fighter: &mut L2CFighterCommon) -> L2CValue {
+    if fighter.is_situation(*SITUATION_KIND_AIR) && VarModule::is_flag(fighter.battle_object, vars::kirby::instance::DISABLE_STONE) {
```
SkewedAskew: You shouldn't need a check for `SITUATION_KIND_AIR` here as the flag gets turned back off if Kirby is grounded anyway

fighters/kirby/src/status/mod.rs
```diff
@@ -31,6 +39,12 @@ unsafe extern "C" fn change_status_callback(fighter: &mut L2CFighterCommon) -> L
         VarModule::set_int(fighter.battle_object, vars::koopa::instance::SPECIAL_N_FIREBALL_COOLDOWN, KOOPA_MAX_COOLDOWN);
     }
 
+    /// Re-enables Stone
+    if fighter.is_situation(*SITUATION_KIND_GROUND) || fighter.is_situation(*SITUATION_KIND_CLIFF) 
+    || fighter.is_status_one_of(&[*FIGHTER_STATUS_KIND_REBIRTH, *FIGHTER_STATUS_KIND_DEAD, *FIGHTER_STATUS_KIND_LANDING, *FIGHTER_STATUS_KIND_GIMMICK_SPRING_JUMP]) {
```
SkewedAskew: Remove the checks for `REBIRTH` and `DEAD` as they already have this flag reset in their statuses below. You also don't need `LANDING` as the `SITUATION_KIND_GROUND` check should cover that

fighters/kirby/src/acmd/aerials.rs
```diff
@@ -145,8 +145,8 @@ unsafe extern "C" fn game_attackairhi(agent: &mut L2CAgentBase) {
...
+        ATTACK(agent, 0, 0, Hash40::new("footr"),  9.0, 30, 120, 0, 5, 4.0, 0.0, -5.6, 0.0, None, None, None, 1.0, 1.0, *ATTACK_SETOFF_KIND_ON, *ATTACK_LR_CHECK_POS, false, 0, 0.0, 0, false, false, false, false, true, *COLLISION_SITUATION_MASK_GA, *COLLISION_CATEGORY_MASK_ALL, *COLLISION_PART_MASK_ALL, false, Hash40::new("collision_attr_normal"), *ATTACK_SOUND_LEVEL_L, *COLLISION_SOUND_ATTR_KICK, *ATTACK_REGION_KICK);
+        ATTACK(agent, 1, 0, Hash40::new("footr"),  9.0, 30, 120, 0, 5, 3.5, 0.0, 0.0, 0.0, None, None, None, 1.0, 1.0, *ATTACK_SETOFF_KIND_ON, *ATTACK_LR_CHECK_POS, false, 0, 0.0, 0, false, false, false, false, true, *COLLISION_SITUATION_MASK_GA, *COLLISION_CATEGORY_MASK_ALL, *COLLISION_PART_MASK_ALL, false, Hash40::new("collision_attr_normal"), *ATTACK_SOUND_LEVEL_L, *COLLISION_SOUND_ATTR_KICK, *ATTACK_REGION_KICK);
```
SkewedAskew: This should be reverted, the sfx_level difference works to audibly convey the difference in properties compared to the early hit

SkewedAskew (changes requested):

fighters/kirby/src/status/mod.rs
jobrien97: We actually do need to check for LANDING as well. Both are included in our other resource refresh checks. There is some weirdness with landing + edge canceling within 1f that won’t always update your situation kind.

fighters/kirby/src/opff.rs
ethangarsia517: Setting the flag in this status results in resetting Down Special usage when used on the ground, or when hitting the ground and transforming back out. It was intended for it to only reset usage when Kirby lands on the ground outside of being in Stone in the same way as how multi-jumps reset...
Unless dev team wants it this way, I can move setting the flag to the ```FIGHTER_KIRBY_STATUS_KIND_STONE_END``` instead.

SkewedAskew: Yeah, go ahead and put it in `STONE_END` main

fighters/kirby/src/status/mod.rs
SkewedAskew: I hate it here

SkewedAskew (approved):

PR #2633, opened 2025-12-18
F1 Grab Tech

SuddyN:
# Experimental!
Grab Tech / Grab Clank:
- Grab Techs are now enabled on the first frame of **standing grabs**
  - *Previously, they were enabled 1F before the grab hitbox*
  - *This change does not apply to non-standard grabs, like tether grabs, or slow direct grabs*
  - *Intended to provide a new defensive option in some scenarios*


SuddyN:
closed to iterate upon in developer environment

PR #2641, opened 2025-12-29
Stale Move Rework and Max DI Correction Levels

BlankMauser:
Note: The visual adjustments and numbers tweaked here are not final and subject to change.

# Purpose and Intent
This pull request is an experimental rework of the stale move queue in smash ultimate. It aims to amplify the existing flavor of HDR's punish game while simultaneously easing the burden of adjusting to higher degrees of DI freedom. The purpose is to take a relatively tedious to engage mechanic and rework it to be more understandable at all skill levels. To complement this, the new stale move queue aims to be simpler and easier to account for during balance adjustments.

## Changes and Additions
In Smash Ultimate attacks you land are stored in a queue and damage scales based on:

1. The amount of times the attack appears in the queue.
2. The position in the queue each attack appears in.

Ult tracks the last 9 attacks you have landed. This includes attacks that are shielded and attacks that you've landed on articles. Stale moves in ultimate also have a considerable difference in damage output compared to games such as Melee, and damage is a large factor in shield stun calculations. You can read specifics [here](https://www.ssbwiki.com/Stale_Moves).

This pull request makes the following changes:

```
(!) Mechanical

- Move staling no longer applies to shields or articles.
- The move stale queue is replaced to be on the defender instead of the attacker.
- The queue now only tracks the last four unique attacks.
- Upon receiving an attack that is already in queue, its placement in queue is refreshed.
- New attacks will replace the oldest attack in queue if the queue is full.
- Getting hit by a stale move will increase your max DI correction by one level. (2 degrees per level up to a max of 19)
- Your max DI correction level is reset upon touching the ground.

[$] Aesthetic

- DI correction vfx lasts one frame longer.
- DI correction vfx color adjusts based on current max DI level.
```

## Technical Breakdown

Projectiles are currently not programmed to be added to the stale move queue. It is possible but currently only fighter states are logged. Maximum DI correction adjusts based on the following table:

| Level | Max Correction | VFX Color |
| ---- | ----- | -------|
| 1 | 15 | Green |
| 2 | 17 | Blue |
| 3 | 19 | Purple |

yuzu_dDiqLuTb80
yuzu_HjM3l28UDE

# Summary

HDR's meta is likely to see a great deal of change soon. The proposal here is meant to be a compromise between its existing punish game while adding new strategic elements and defensive agency. DI affects everything from kill confirms to punishing crouch cancelling. This rework gives testers a more consistent reference point for lower vs. higher degrees of freedom, makes shieldstun more consistent to balance, and brings developers another knob to tweak balance without completely reframing the current state of the game.

I believe this helps HDR play to its current identity of having a high punish game while passively offering more defensive options against repetition. It is a fresh take compared to other smash titles while not being invasive to the kits of potentially weaker characters. Players can easily keep track of their stale moves because only the last four unique attacks are logged. This makes it easier for them to actively use fresh or stale attacks and intentionally force different DI mix-ups and situations for their opponent.


SuddyN:
We're not interested in tying this mechanic to Directional Influence, but the reimplementation behind the stale move queue is great, and we may want to use it in the future. For now I've approved the build action for this PR so it can be tested in the launcher, but I would appreciate if you could open a new PR with the StaleMoveQueue implementation including contsts, boma extensions, vars, etc., but excluding the actual implementation of DI param-hooking and DI-line coloring. We will include that work, and the team is actively discussing internally how we might use it.

SuddyN:
closed as not planned

PR #2642, opened 2025-12-29
Stupid Gordo Texture Bugfix

SirTanknSpank:
Fixes the (surprisingly common) bug where gordo loses its texture due to some weird funky gordo status shenanigans. 

Why is this projectile like this?


SuddyN (approved):
lgtm if it works

SkewedAskew:
Can't wait for it to somehow break more at a random point in the future

PR #2646, opened 2026-01-08
Add Baby Dashing

SuddyN:
Adds Baby Dashing as a mechanic, reimplemented for HDR
- on F2 of Dash, if the stick is below the dash threshold but not backwards, transition into the Run Brake status
- Baby Dash speed is parameterized per-character, and is currently set to 1.125

For most characters, this travels much less far than Pivot or Perfect pivot. However, since the Baby Dash speed is parameterized separate from the Dash Speed, and Perfect Pivot has a speed has a speed multiplier of 0.75x, characters with lower than a 1.5 dash speed can travel faster with Baby Dash than with Perfect Pivot.

Baby Dash also offers unique advantages, such as using grounded attacks forward without turning around, or micro-positioning in neutral.

Unlike Pivot and Perfect Pivot, Baby Dash cannot be used while Dash Dancing. This restriction arose as a limitation of preserving the Pivot and Perfect Pivot input clarity. This fact limits Baby Dash's usefulness in increasing burst range.

Currently, Baby Dash can only be performed on F2 of your Dash animation. This restricts the input leniency significantly, and is intended to (re)introduce an impactful skill barrier. There may be a concern about input inconsistencies on some controllers, but it's possible to play around polling inconsistencies; by flicking the stick slightly up/down diagonally to initiate your dash, and then rolling the stick up/down (away from the X axis), it's possible to consistently perform a Baby Dash even with an imperfect controller. This is possible because the same horizontal dash threshold is used both for starting the dash and for checking whether Baby Dash is valid.

This PR is intended for testing purposes, and has not passed a Development Team design or balance review. Please playtest this PR with other people before forming an opinion, and remember to have fun :)


SuddyN:
not being considered at this time

PR #2647, opened 2026-01-09
Stutter step + perfect pivot adjustments

BlankMauser:
These are adjustments made to a [hypothetical implementation of baby dash.](https://github.com/HDR-Development/HewDraw-Remix/pull/2646) It reworks it to be closer to a grounded stutter step while giving it different tradeoffs from perfect pivot.

### Frame 4 and 3 of dash can be cancelled into grounded tilts/smashes and specials while retaining a set momentum.

Importantly this does not include jab or grab. Certain grounded specials can also be b-reversed while retaining said momentum.

### Has a per character speed multiplier that can be adjusted.

Currently the multiplier is set to 0.925 globally. This param labelled as "stutter_step_speed_mul" can be adjusted per character.

### Has a 1fr window for max distance, and a 2fr window for regular distance.

This is mostly an added skill expression window but the regular distance window being double means players can expect consistent distance most of the time.

The pull request has a small experimental tweak to perfect pivot to make it more consistent with imperfect controllers. By tapping or rolling diagonally you can get perfect pivot without relying on analog polling to go back to neutral.


SuddyN:
not being considered at this time

PR #2668, opened 2026-05-07
Finish ssbusync integration.

BlankMauser:
Symbaker no longer needed. Used a custom macro from https://github.com/BlankMauser/smash-ultelier/blob/master/crates/ultelier-macros/src/lib.rs instead to avoid symbol conflicts.

Buffer mode is chosen at CSS based on player count. See player_port.rs for implementation.

Use install_menu_hooks() instead of install_auto_profile_switcher() to run without custom overclocker:

```rust
    //*Install only menu hooks to run without custom overclocker.
    //sync::runtime::install_menu_hooks();
    
    sync::runtime::install_auto_profile_switcher();
```

This does not include modified vfx assets.


SuddyN:
pursuing a different method of integration with HDR

PR #2711, opened 2026-08-10
Implement short hop aerial macro in CSS menu

EIREXE:
Reimplements the previously commented out short hop aerial macro from vanilla ultimate and makes it optionally toggleable from within the small menu in the CSS.

Some work had to be done in `lua.rs` because otherwise the bits indicating the macro was enabled would get cleared on when saving from the vanilla options menu.

### DIFF
--- a/dynamic/src/ext.rs
+++ b/dynamic/src/ext.rs
@@ -328,6 +328,7 @@ bitflags! {
         const CStickOverride = 0x200000;
         const RivalsWallJump = 0x400000;
         const TreadJump = 0x800000;
+        const ShortHopAerialMacro = 0x1000000;
 
         const SpecialAll  = 0x20802;
         const AttackAll   = 0x201;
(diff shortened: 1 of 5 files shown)


SuddyN (commented):
This code looks good. However, if memory serves, we previously had an issue with the size of the buttons struct we were able to use online, due to networking constraints. The specific issue is unfortunately undocumented institutional knowledge, but I'll do some research and ask around. In the meantime, please test this online on official hardware, and document any crashes or desyncs.

EIREXE:
> This code looks good. However, if memory serves, we previously had an issue with the size of the buttons struct we were able to use online, due to networking constraints. The specific issue is unfortunately undocumented institutional knowledge, but I'll do some research and ask around. In the meantime, please test this online on official hardware, and document any crashes or desyncs.

I do not have my switch on hand (I'm on vacation) if anyone can try it please be my guest.

